PR #1, opened 2021-01-18
Added safe handling for some nested field accesses that were causing …

craigwblake:
…errors.


curatingbits (approved):
safeGaurd looks good. Thank you.

PR #38, opened 2021-04-30
Changing project nickname for eternal pumps

CosineXArt:
Making artbot in the #dmitri-cherniak channel respond to both "pump" and "eternal pump".


jakerockland (commented):
LGTM!

PR #40, opened 2021-05-04
Add Giveaway Functionality to the ArtBot

CosineXArt:
Uses the node package discord-giveaways to manage random drawings on Discord.

https://github.com/Androz2091/discord-giveaways

You will need to add CHANNEL_MINTS to the .env file, and give someone the "giveaway master" role to be able to kick off giveaways.  It will silently ignore anything that doesn't start with "giveaway!<cmd>", so I think it's not likely to accidentally start an event, but we can make it more complicated if you like.

Only thing not done yet is I'd like to play with using custom Emojis, but we'd need to get those set up on the test server first.


index.js
```diff
@@ -93,6 +95,29 @@ app.listen(PORT, function() {
 // Bot setup.
 const bot = new Client();
 
+
+//Manage Giveaways with Artbot
```
jakerockland, line 99: Minor thing, but can we move this new chunk of code to live after the:

```
bot.login(TOKEN);

bot.on("ready", () => {
  console.info(`Logged in as ${bot.user.tag}!`);
});
```

I think its a bit easier to visually parse this file that way.

Utils/giveawayCommands.js
```diff
@@ -0,0 +1,133 @@
...
+            return message.channel.send('Unable to find a giveaway for `'+ args.join(' ') +'`.');
+        }
+
+        // Reroll the giveaway
+        bot.giveawaysManager.reroll(giveaway.messageID)
+        .then(() => {
+                   // Success message
+                   msg.channel.send('Giveaway rerolled!');
+        })
+        .catch((e) => {
```
jakerockland, line 122: Nit: here and in the few places above can we use `catch((error) =>` rather than `catch((e) =>` as a pattern? It makes the code that follows a bit easier to read (and code is read more often than written).

Utils/giveawayCommands.js
```diff
@@ -0,0 +1,133 @@
...
+        let giveaway = 
+            // Search with giveaway prize
+            bot.giveawaysManager.giveaways.find((g) => g.prize === args.join(' ')) ||
+            // Search with giveaway ID
+            bot.giveawaysManager.giveaways.find((g) => g.messageID === args[0]);
+
+        // If no giveaway was found
+        if(!giveaway){
+            return message.channel.send('Unable to find a giveaway for `'+ args.join(' ') +'`.');
+        }
```
jakerockland, line 114: This logic is duplicated in a couple places, can we please split off into its own little helper function?

Utils/giveawayCommands.js
```diff
@@ -0,0 +1,133 @@
...
+    const ms = require('ms'); // npm install ms
+    const args = msg.content.slice('giveaway!'.length).trim().split(/ +/g);
+    const command = args.shift().toLowerCase();
+
+    // giveaway!start 2d 1 Awesome prize!
+    // Will create a giveaway with a duration of two days, with one winner and the prize will be "Awesome prize!"
+    
+
+    // If this is a valid command, and the member doesn't have enough permissions, return an error
+    if ((command === "start") || (command === "reroll") || (command === "end")) {
```
jakerockland, line 17: In addition to "start", "reroll" and "end", could we please add a "help" command that explains how to use the giveaway function? I think inevitably there will be someone who needs to be reminded, is taking over the giveaway-runner role, etc.

Utils/giveawayCommands.js
```diff
@@ -0,0 +1,133 @@
...
+        if(!msg.member.hasPermission('MANAGE_MESSAGES') && !msg.member.roles.cache.some((r) => r.name === "giveaway master")){
+            return msg.channel.send(':x: You need to have the right permissions to start giveaways.');
+        }
+    }
+
+    if (command === "start") {
+
```
jakerockland, line 24: I think the logic within this `if (command === "start")` branch is big enough to be its own helper function.

I think we could preferably do this for each of the commands (start, reroll, end, and help), make them helpers that are internal to this file, which would allow the if-else branching here to be a bit easier to parse over and then one can dig into each helper method as needed.

jakerockland (changes requested):
This is fantastic work 💯 

I've added couple requests around a bit of refactoring and adding a "help" command and then I think we should be set to push this live.

index.js
CosineXArt: Let me test that out, I thought I had to add these items before the login command, but I'll check if that's true.

Utils/giveawayCommands.js
CosineXArt: Sure thing!

Utils/giveawayCommands.js
CosineXArt: Good point, will do.

Utils/giveawayCommands.js
CosineXArt: Excellent point, I'll add that.

Utils/giveawayCommands.js
CosineXArt: Sure.

index.js
CosineXArt: Looks like I can do it this way.

CosineXArt:
All your comments should be addressed with this latest commit.

Utils/giveawayCommands.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    if (command === "reroll") {
+        return handleGiveawayReroll(msg, bot, args);
+    }
+   
+    if (command === "help") {
+        return msg.channel.send("Giveaway Usage:  giveaway![start|end|reroll] <time> <winner_count> <prize>");
+    }
+}
```
jakerockland, line 149: One minor nit here: somewhat of a style preference, I think its a bit easier to read if this branching uses if-else rather than 4 if statements as it providers a better guarantee to the reader that only one of these may be executed per command.

jakerockland (commented):
Just one minor thing here that would be nice if you can tweak. Non-blocking though if its a hassle.

I'll merge and push this tonight :) looks great!! Thank you.

jakerockland (approved):
Looks great, thank you!! 😄

PR #41, opened 2021-05-07
Setting up the Jason Ting Channel Bot

CosineXArt:
Added Bubble Blobby and Light Beams project bots and added them to the Jason Ting curated artist channel.


jakerockland (approved):
This is great, but I'm going to remove light beams for now before pushing live. We're currently only reflecting Curated and Playground with the bots and until we have a good consistent story for supporting factory projects I want to hold off on adding support there :)

PR #42, opened 2021-05-12
Teach the artbot to tell us the gas price

CosineXArt:
Just a simple wrapper around the ethgasstation.info API.  Open to suggestions on how it should be invoked.  I also added some decoration for super-high gas prices, if you'd like to remove that, I'm not attached to it.

Requires an API key, which can be obtained at https://data.defipulse.com/
added to .env as
GASSTATION_API_KEY=<key>

### DIFF
--- a/Utils/smartBotResponse.js
+++ b/Utils/smartBotResponse.js
@@ -91,9 +91,9 @@ async generateGasPriceMessage() {
     let gasStationResponse = await fetch(`https://ethgasstation.info/api/ethgasAPI.json?api-key=${GASSTATION_API_KEY}`);
     let gasStationData = await gasStationResponse.json();
     let fireString = '';
-    if (gasStationData.average > 400) {
+    if (gasStationData.average > 4000) {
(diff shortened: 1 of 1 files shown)


curatingbits:
Is ethgasstation.info better than web3.js? I believe web3 can get the current eth gas price. I will default to Rockland on this. Thanks for your help/work CosineXArt! Good job.

CosineXArt:
> Is ethgasstation.info better than web3.js? I believe web3 can get the current eth gas price. I will default to Rockland on this. Thanks for your help/work CosineXArt! Good job.

Good question!  I'll take a look at what web3.js offers, this was just one that looked pretty straightforward to implement.

jakerockland:
It looks like web3js only exposes the average gas price but not the degree of nuance that ETHGasStation's API provides. Given the API has a pretty generous free tier I say we go with this approach for now :)

jakerockland:
Thank you @CosineXArt, great stuff 💯

PR #54, opened 2021-07-25
first draft of factory bot

CosineXArt:
This is a draft.  Code's still a little messy, but it now covers the main cases, handles projects that aren't completely minted, and refreshes the metadata on an interval.  (It takes less than a minute on average to refresh, and updates incrementally, so it shouldn't be noticeable by the user).

Add the following two .env variables to make it work:
CHANNEL_FACTORY
METADATA_REFRESH_INTERVAL_MINUTES

### DIFF
--- a/Classes/FactoryBot.js
+++ b/Classes/FactoryBot.js
@@ -11,11 +11,9 @@ const ProjectBot = require("./ProjectBot").ProjectBot;
 
 const web3 = new Web3(Web3.givenProvider || "ws://localhost:8545");
 
-const MINT_ADDRESS = "0x0000000000000000000000000000000000000000";
 const V2_MINTING_CONTRACT_ADDRESS = "0xa7d8d9ef8d8ce8992df33d8b8cf4aebabd5bd270";
-const EMBED_COLOR = 0xff0000;
-const UNKNOWN_ADDRESS = "unknown";
-const UNKNOWN_USERNAME = "unknown";
+
+// Refresh takes around one minute, so recommend setting this to 60 minutes
 const METADATA_REFRESH = process.env.METADATA_REFRESH_INTERVAL_MINUTES;
 
 // This array will hold ProjectBot classes for all the Factory Projects we find
--- a/Classes/FactoryBot.js
+++ b/Classes/FactoryBot.js
@@ -18,7 +18,7 @@ const UNKNOWN_ADDRESS = "unknown";
 const UNKNOWN_USERNAME = "unknown";
 const METADATA_REFRESH = process.env.METADATA_REFRESH_INTERVAL_MINUTES;
 
-
+// This array will hold ProjectBot classes for all the Factory Projects we find
 let factoryBotList = [];
 
 class FactoryBot {
@@ -34,8 +34,8 @@ class FactoryBot {
     for (let i = 0; i < projectList.length; i++) {
        let projectData = await getArtBlocksProject(projectList[i]); 
 
-       console.log(`Refreshing project cache for Project ${projectList[i]} ${projectData.name}`);
-       if (isFactoryProject(projectList[i])) {
+       if (projectData && isFactoryProject(projectList[i])) {
+	   console.log(`Refreshing project cache for Project ${projectList[i]} ${projectData.name}`);
            let newBot = new ProjectBot(
                    projectList[i]*1000000,
                    V2_MINTING_CONTRACT_ADDRESS,
@@ -52,7 +52,6 @@ class FactoryBot {
  
   async handleNumberMessage(msg) {
 
-    console.log(msg.content);
     let content = msg.content;
 
     if (content.length <= 1) {
@@ -62,13 +61,12 @@ class FactoryBot {
       return;
     }
 
-    let detailsRequested = content.toLowerCase().includes("detail");
     let afterTheHash = content.substring(1);
     let nameIndex = content.substr(content.indexOf(" ")+1).toLowerCase().replace(/\s+/g, '');
-    console.log(nameIndex);
-    if (factoryBotList[nameIndex]) {
-       console.log(factoryBotList[nameIndex]);
-       factoryBotList[nameIndex].handleNumberMessage(msg);
+    console.log(`Searching for project ${nameIndex}`);
+    let projBot = factoryBotList[nameIndex];
+    if (projBot) {
+       projBot.handleNumberMessage(msg);
     }
   }
 
--- a/Classes/FactoryBot.js
+++ b/Classes/FactoryBot.js
@@ -1,0 +1,77 @@
+require("dotenv").config();
+const {
+  MessageEmbed
+} = require("discord.js");
+const fetch = require("node-fetch");
+const Web3 = require("web3");
+const getArtBlocksPlatform = require("../Utils/parseArtBlocksAPI").getArtBlocksPlatform;
+const getArtBlocksProject = require("../Utils/parseArtBlocksAPI").getArtBlocksProject;
+const isFactoryProject = require("../Utils/parseArtBlocksAPI").isFactoryProject;
+const ProjectBot = require("./ProjectBot").ProjectBot;
+
+const web3 = new Web3(Web3.givenProvider || "ws://localhost:8545");
+
+const MINT_ADDRESS = "0x0000000000000000000000000000000000000000";
+const V2_MINTING_CONTRACT_ADDRESS = "0xa7d8d9ef8d8ce8992df33d8b8cf4aebabd5bd270";
+const EMBED_COLOR = 0xff0000;
+const UNKNOWN_ADDRESS = "unknown";
+const UNKNOWN_USERNAME = "unknown";
+const METADATA_REFRESH = process.env.METADATA_REFRESH_INTERVAL_MINUTES;
+
+
+let factoryBotList = [];
+
+class FactoryBot {
+  constructor(projectNumber, mintContract, editionNumber, projectName) {
+    this.initialize();
+    setInterval(this.initialize, METADATA_REFRESH * 60000);
+  }
+
+  async initialize() {
+
+    let projectList = await getArtBlocksPlatform();
+
+    for (let i = 0; i < projectList.length; i++) {
+       let projectData = await getArtBlocksProject(projectList[i]); 
+
+       console.log(`Refreshing project cache for Project ${projectList[i]} ${projectData.name}`);
+       if (isFactoryProject(projectList[i])) {
+           let newBot = new ProjectBot(
+                   projectList[i]*1000000,
+                   V2_MINTING_CONTRACT_ADDRESS,
+                   projectData.invocations,
+                   projectData.name
+                   );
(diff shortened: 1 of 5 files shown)


Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+        if (node.tagName === 'pre') return;
+
+        keyValue = node.text.split(':');
+        if (keyValue.length >= 2 && keyValue[1]) {
+            platform[keyValue.shift().split(' ').join('_').toLowerCase()] = keyValue.join(':').trim();
+        }
+    });
+
+    platform.projects_list = platform.projects_list.split(',').map((p) => parseInt(p));
+        return platform.projects_list;
```
jakerockland, line 25: Nit: Can you fix alignment of this return statement please?

Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+        delete project.http;
+        delete project.https;
+    }
+
+    project.token_ids = [];
+    for (let i = 0; i < project.invocations; i++) {
+        project.token_ids.push(id * 1000000 + i);
+    }
+
+        return project;
```
jakerockland, line 59: Nit: Can you fix alignment of this return statement please?

Utils/parseArtBlocksAPI.js
jakerockland, line 25: Nit: Can you fix alignment of this return statement please?

Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+        if (keyValue.length >= 2 && keyValue[1]) {
+            project[keyValue.shift().split(' ').join('_').toLowerCase()] = keyValue.join(':').trim();
+        }
+    });
+
+    if (project.http || project.https) {
+        project.website = `http${project.https ? `s:${project.https}` : `:${project.http}`}`;
+        delete project.http;
+        delete project.https;
+    }
```
jakerockland, line 52: Could you add a comment please explaining why this needs to be done?

Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+
+    project.token_ids = [];
+    for (let i = 0; i < project.invocations; i++) {
+        project.token_ids.push(id * 1000000 + i);
```
jakerockland, line 56: We should probably define 1000000 as a constant at the top of the file here, this seems easy to typo if someone was modifying this in the future. 

Also reminds that that I should do some minor refactoring of the ProjectBot class in the future, though that is clearly out of scope here for this PR.

Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+
+        return project;
+};
+
+async function isFactoryProject (id) {
+
+       let tokenid = id*1000000;
+       let is_factory = false;
+       try {
+           const response = await fetch(`https://api.artblocks.io/token/${tokenid}`, { timeout: 5000 });
```
jakerockland, line 67: We probably should be using timeouts in our other fetches like you are here, but curious why you explicitly added the timeout (were you seeing odd behavior without it)?

Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+    for (let i = 0; i < project.invocations; i++) {
+        project.token_ids.push(id * 1000000 + i);
+    }
+
+        return project;
+};
+
+async function isFactoryProject (id) {
+
+       let tokenid = id*1000000;
```
jakerockland, line 64: Nit: I think tokenID is slightly more readable.

Utils/parseArtBlocksAPI.js
```diff
@@ -0,0 +1,83 @@
...
+       let is_factory = false;
```
jakerockland, line 65: Nit: generally we're using camelCase elsewhere, could we keep that the case here with is_factory->isFactory?

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+const fetch = require("node-fetch");
+const Web3 = require("web3");
+const getArtBlocksPlatform = require("../Utils/parseArtBlocksAPI").getArtBlocksPlatform;
+const getArtBlocksProject = require("../Utils/parseArtBlocksAPI").getArtBlocksProject;
+const isFactoryProject = require("../Utils/parseArtBlocksAPI").isFactoryProject;
+const ProjectBot = require("./ProjectBot").ProjectBot;
+
+const web3 = new Web3(Web3.givenProvider || "ws://localhost:8545");
+
+const MINT_ADDRESS = "0x0000000000000000000000000000000000000000";
```
jakerockland, line 14: Is this MINT_ADDRESS needed or a reference that has since been cleaned up?

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+const V2_MINTING_CONTRACT_ADDRESS = "0xa7d8d9ef8d8ce8992df33d8b8cf4aebabd5bd270";
+const EMBED_COLOR = 0xff0000;
+const UNKNOWN_ADDRESS = "unknown";
+const UNKNOWN_USERNAME = "unknown";
```
jakerockland, line 18: Similarly, are these needed?

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+const METADATA_REFRESH = process.env.METADATA_REFRESH_INTERVAL_MINUTES;
```
jakerockland, line 19: What do you think an appropriate value is here? My inclination would be to start with something like 3 hours or 6 hours and work down from there if needed at all? Perhaps we could do less though, like 1 hour. There aren't that many API calls involved.

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+       let projectData = await getArtBlocksProject(projectList[i]); 
+
+       console.log(`Refreshing project cache for Project ${projectList[i]} ${projectData.name}`);
+       if (isFactoryProject(projectList[i])) {
+           let newBot = new ProjectBot(
+                   projectList[i]*1000000,
+                   V2_MINTING_CONTRACT_ADDRESS,
+                   projectData.invocations,
+                   projectData.name
+                   );
```
jakerockland, line 44: Love that we can do this. 💜🙏

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+    let content = msg.content;
+
+    if (content.length <= 1) {
+      msg.channel.send(
+        `Invalid format, enter # followed by the piece number of interest.`
+      );
+      return;
+    }
+
+    let detailsRequested = content.toLowerCase().includes("detail");
```
jakerockland, line 65: Do we need to pull this out as a var?

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+
+const MINT_ADDRESS = "0x0000000000000000000000000000000000000000";
+const V2_MINTING_CONTRACT_ADDRESS = "0xa7d8d9ef8d8ce8992df33d8b8cf4aebabd5bd270";
+const EMBED_COLOR = 0xff0000;
+const UNKNOWN_ADDRESS = "unknown";
+const UNKNOWN_USERNAME = "unknown";
+const METADATA_REFRESH = process.env.METADATA_REFRESH_INTERVAL_MINUTES;
+
+
+let factoryBotList = [];
```
jakerockland, line 22: Would be nice to have a small comment explaining what gets stored in this variable (that it is a place in memory for the project handler classes to live).

Classes/FactoryBot.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    let detailsRequested = content.toLowerCase().includes("detail");
+    let afterTheHash = content.substring(1);
+    let nameIndex = content.substr(content.indexOf(" ")+1).toLowerCase().replace(/\s+/g, '');
+    console.log(nameIndex);
+    if (factoryBotList[nameIndex]) {
+       console.log(factoryBotList[nameIndex]);
+       factoryBotList[nameIndex].handleNumberMessage(msg);
+    }
```
jakerockland, line 72: Nit: we could probably factor out `factoryBotList[nameIndex]` to simplify a bit here.

jakerockland (changes requested):
This generally looks really really great! A couple minor nits that are appreciated and not necessary as well as some various minor questions 💯 Thank you @CosineXArt 🙏

Utils/parseArtBlocksAPI.js
CosineXArt: absolutely

Utils/parseArtBlocksAPI.js
CosineXArt: done, although I wonder if it would be easier to just change this to 1e6 to avoid the typo problem.

Utils/parseArtBlocksAPI.js
CosineXArt: Yes, if you hit the ArtBlocks API with ~3-400 requests a few of them fail, and without the timeout, thing error unpredictably (closed sockets, etc.).  But now that you point this out, I need a timeout on my fetches of the project and platform endpoints as well.

Utils/parseArtBlocksAPI.js
CosineXArt: I think this was a copy paste error, and I don't think we need it.  Deleted.

Classes/FactoryBot.js
CosineXArt: I ran it for a while on 5 minute refreshes in test, and it was fine.  It takes a minute or so to complete the refresh, so 1 hour would probably be fine.

Classes/FactoryBot.js
CosineXArt: LOL, we finally fix the problem of the randomizer and unfinished mints, and now we don't have any more uncompleted projects.  At least you can #? squiggles!

Classes/FactoryBot.js
CosineXArt: nope.  Accidental copy/paste error from the ProjectBot class.

Classes/FactoryBot.js
CosineXArt: I can add one.

Classes/FactoryBot.js
CosineXArt: sure thing.

Classes/FactoryBot.js
jakerockland: Rad, that sounds good to me! Would you mind adding a comment to this effect?

Utils/parseArtBlocksAPI.js
jakerockland: Aha, thanks for clarifying :) If you don't mind, could you add a comment to this effect?

jakerockland (approved):
This looks great! Great work and thank you for doing this. 💜🙏

PR #62, opened 2021-08-12
add metamask high gas explanation

conundrumer:
During drops, prospective minters frequently complain about extremely high gas, and we have to explain over and over again that it's metamask doing it. Now we can explain with artbot by doing `@artbot high gas?` or `@artbot gas high?`


jakerockland:
This is fantastic, thank you! :)

Issue #99, opened 2021-10-01
Fix syntax errors introduced in PR #98


ryley-o:
artbot currently doesn't run on master. Good example of why we want to add CI.

PR #100, opened 2021-10-01
Fixes issue #99

ryley-o:
minor edits to get artbot working again


aaronpenne:
ty for the catch, I need to set up a local test env

PR #106, opened 2021-10-05
Refactor channel ENV variables to be configured via JSON

### LINKED ISSUE #83: Refactor channel ENV variables to be configured via JSON
With the number of channels we're adding, we will eventually run into constraints with the # of bytes we're trying to store in ENV variables on Heroku.

We eventually will need to move to using a different configuration system for channels, for which I would suggest using JSON.

This could be as simple as something like `channels.json` that looks like:

```
{
  "CHANNEL_FIDENZA_SALES": 880937060075192320
  "CHANNEL_FIDENZA_GENERAL": 879488624197001293
  // ... etc
}
```

Or we could (this is preferable but more nuanced and this work could be done in two parts as such) configure the entire project bots (each `ProjectBot`) in JSON which would make it easier for to add support for new artists/projects. 

E.g., something like:

```
{
  "fidenza": {"channel": 880937060075192320, projectNumber: 78, name: "Fidenza", mintCount: 999}
  // ... etc
}
```

ryley-o:
Closes #83.

- Projects bots and named mappings are now all defined in ProjectConfig/projectBots*.json
- Channels and associated project bot trigger words and/or tokenIDs are now all defined in ProjectConfig/projectBots*.json
- Minter contracts defined in minterContracts.json

Environment variables are now only required for items that should be configured at runtime, or are non-public. Including env.example file to describe remaining recommended set of environment variables.


.env.example
```diff
@@ -0,0 +1,5 @@
+ARTBOT_IS_PROD=false;
+TOKEN=<contact maintainer for artbot-jr token is desired>
+METADATA_REFRESH_INTERVAL_MINUTES=60
+# CHANNEL_ADDRESS_COLLECTION=
```
jakerockland, line 4: Can we just remove this from the example rather than comment it out?

.env.example
```diff
@@ -0,0 +1,5 @@
+ARTBOT_IS_PROD=false;
```
jakerockland, line 1: Can we just remove this line if it is false? I'm worried that including it at all may lead to some people accidentally flipping it to true to "see what happens".

ProjectConfig/minterContracts.json
```diff
@@ -0,0 +1,4 @@
+{
```
jakerockland, line 1: The original mistake here was mine in calling these minter contracts, that was a misnomer. These are actually the "Core" contracts. Can we please rename this file and associated usage to `coreContracts` instead for better naming consistency with the rest of our codebase? Apologies for the churn.

Classes/ProjectBot.js
```diff
@@ -28,6 +36,11 @@ class ProjectBot {
       return;
     }
 
+    // decode any mappings
+    if (this.namedMappings) {
+      content = this.namedMappings.transform(content)
```
jakerockland, line 41: Nit: can you add a semicolon at the end of the line here please?

ProjectConfig/projectConfig.js
```diff
@@ -0,0 +1,131 @@
...
+
+  static buildChannelHandlers(ChannelsJson) {
+    const channels = {};
+    Object.entries(ChannelsJson).forEach(([chID, chParams]) => {
+      channels[chID] = new Channel(chParams);
+    });
+    return channels;
+  }
+
+  static buildChIdByName(channels) {
```
jakerockland, line 105: Nit: I don't think we need to abbreviate here, can we update this to `buildChannelIDByName` please?

ProjectConfig/projectConfig.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+    const botName =
+      channel.botNameFromNumberMsgContent(msg.content.toLowerCase());
+    if (botName === null) {
+      // should never occur for anything we are being asked to route
+      console.error(`Channel ID: ${channelID} does not have a ProjectBot`);
+    }
+    this.projectBots[botName].handleNumberMessage(msg);
+  }
+}
```
jakerockland, line 128: Do you mind adding short documentation comments for each of these functions please?

jakerockland (changes requested):
This is excellent work, thank you @ryley-o 🥳 

A couple minor requests here inline, and then one slightly broader request: Do you mind adding a new markdown file, named something like "JSON_SCHEMA.md", which gives a very brief explanation for the valid schema for each of the new JSON config files being introduced?

ProjectConfig/projectBots_dev.json
```diff
@@ -0,0 +1,34 @@
+{
+    "squiggleBot": {
+        "projectNumber": 0,
+        "mintContract": "OG",
+        "editionNumber": 10000,
```
jakerockland, line 5: One more naming preference here, can we use "editionSize" or "editionCount" rather than "editionNumber"?

.env.example
ryley-o: Yes

.env.example
jakerockland: This part is already set in Prod 🥳 once this PR is merged I will go through and clean up old ENV variables, as we should only need a very few set now in Heroku after this merges, and will be careful to not delete this 😂

ProjectConfig/minterContracts.json
ryley-o: Yes

jakerockland:
TY for the updates @ryley-o ! This LGTM, pending on question, re: the JSON_SCHEMA.md or something of the like, did you write that and forget to add to the commit, or are you planning to follow up on that via a later PR? If the later, that is fine, but can you file a bug to capture this please?

Issue #104, opened 2021-10-04
Add basic unit test coverage for new projects


jakerockland:
Ideally bare bones here is test coverage for the addition of new project support, and then we can branch out from there in additional coverage down the line.

PR #116, opened 2021-10-12
Add missing skulptuurBot definition to projectBots json

ryley-o:
SkulptuurBot definition was unintentionally left off during refactor. This type of error will not occur once issue #104 is completed.


jakerockland:
Whoops, rad thanks for fixing @ryley-o and thanks for flagging @mathewzilla

PR #119, opened 2021-10-12
Fix global replace to implement Owner/Winner -> Seller/Buyer

ryley-o:
Parenthesis in regex were causing artbot to not replace Owner and Winer with Seller and Buyer, respectively. Also, updated to only globally replace Winner since Owner will be at the start of description, and that is all that we want to replace.


ryley-o:
@jakerockland I missed that the regex including parenthesis in PR #118, so updating the regex to match Owner and Winner for this PR.

No way to test triager locally at the moment, but I'm confident that this will run fine!

PR #141, opened 2021-11-12
Added Parade and 24 Heures

jrmcdaniel:
Only issue is both projects are still open. I don't think this will be a problem for sharing specific mints, but will probably not work well with the randomBot. Seeing as 24 Heures may be open a long time, I think it's worth it as there is a ton of sharing in Alexis' channel.

### DIFF
--- a/ProjectConfig/channels.json
+++ b/ProjectConfig/channels.json
@@ -407,6 +407,11 @@
         "name": "loren-bednar",
         "projectBotHandlers": {
             "default": "phasesBot"
+            "stringTriggers": {
+                "paradeBot": [
+                    "parade"
+                ]
+            }
         }
     },
     "882129221659537440": {
--- a/ProjectConfig/channels.json
+++ b/ProjectConfig/channels.json
@@ -228,6 +228,14 @@
                 "obiceraBot": [
                     "obicera"
                 ]
+                "heuresBot": [
+                    "heures",
+                    "hours",
+                    "24heures",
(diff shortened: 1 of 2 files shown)


ProjectConfig/channels.json
```diff
@@ -407,6 +415,11 @@
         "name": "loren-bednar",
         "projectBotHandlers": {
             "default": "phasesBot"
```
ryley-o, line 417: can you add a comma so this becomes: `"default": "phasesBot",`

ProjectConfig/channels.json
```diff
@@ -228,6 +228,14 @@
                 "obiceraBot": [
                     "obicera"
                 ]
```
ryley-o, line 230: can you add a comma so this becomes `],`

ryley-o (changes requested):
@jrmcdaniel See my comments for a couple minor requested json-formatting changes.
Later today, I'll also check and see what random will do with the unfinished project. I think it might be quirky, but as long as it doesn't break things, should be fine :)
Thanks!

ryley-o:
Thanks!

When asking for a random `#?` token from a project that is not yet fully minted, artbot currently may return "undefined" if it randomly chooses a token that has not yet been minted. I'll add an issue to update that logic because our project api does return `invocations`, which we could be using instead of `Maximum Invocations`.

ryley-o (approved):

PR #155, opened 2021-11-26
Dive support added

jrmcdaniel:
If the previous crash was due to something I'm doing and not due to OS webhooks, let me know and I'll put a pause on adding the new projects.

### DIFF
--- a/ProjectConfig/channels.json
+++ b/ProjectConfig/channels.json
@@ -460,6 +460,12 @@
             "default": "autologyBot"
         }
     },
+    "912771594357714964": {
+        "name": "ippsketch",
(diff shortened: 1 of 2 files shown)


jakerockland:
Closes #154 

@ryley-o would you PTAL on reviewing this one?

jrmcdaniel:
Updated with Bent as well.

ryley-o:
Awesome, thanks! also, @jrmcdaniel nothing you did last time would have caused artbot to break that I know of.

I can take a look at this on 11/30 (in addition to a bunch of other artbot items) 😄

ryley-o (approved):
LGTM, Thanks!

PR #157, opened 2021-12-01
Add asemica and saturzione bots

### LINKED ISSUE #152: Re-add Asemica and Saturazione from PR #146
PR #146 added support for Asemica and Saturazione, but it was reverted due to artbot locking up and crashing.

Re-add support for those projects. Determine if the added projects were what caused artbot to lock up, or if the issue was related to the OpenSea webhooks which were also down when the prior artbot crash occurred.

ryley-o:
closes #152

### DIFF
--- a/ProjectConfig/channels.json
+++ b/ProjectConfig/channels.json
@@ -376,6 +376,9 @@
             "stringTriggers": {
                 "rinascitaBot": [
                     "rina"
+                ],
+                "saturazioneBot": [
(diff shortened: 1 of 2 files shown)


ryley-o:
No reason this should fail, previous issues may have been due to Heroku watching master instead of main branch, as well as OpenSea webhooks failing. Will watch logs and test after deployment to see if any artbot issues arise.

PR #171, opened 2021-12-15
Updated projects

jrmcdaniel:
Added a few of the latest curated and playgrounds.


jrmcdaniel:
Added Wrapture with a maximum set at 50. If it opens to 666 (there can't be anything inbetween), we can update the max project size. Until we get the change allowing for random selection of partially minted projects, this is probably the most straight forward way to allow for random calls.

mathewzilla (commented):
came here to make this change myself! LGTM but I don't decide these things

jakerockland (approved):
LGTM as well :)

PR #173, opened 2021-12-18
Add IC Support in Tyler Hobbs Server

thobbs:
Adds the icBot to the Tyler Hobbs server. Also updates the channel name from "fidenza" to "the-place-to-be", which is the new name of the main channel.

Please let me know if anything looks incorrect here! Thanks for your work :)

### DIFF
--- a/ProjectConfig/channels.json
+++ b/ProjectConfig/channels.json
@@ -366,9 +366,16 @@
         }
     },
     "879488624197001293": {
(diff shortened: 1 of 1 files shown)


jakerockland:
This LGTM, though note that I believe due to the current OS API outage artbot is temporarily impacted.

jakerockland (approved):

PR #185, opened 2022-01-13
Fix incomplete-control channel ID

jonkensta:
The one published previously is not correct, and this change sets it to the correct one.


jakerockland (approved):
LGTM

PR #200, opened 2022-01-31
Make ArtBot resilient to OpenSea being down

### LINKED ISSUE #190: Handle OpenSea API failures more gracefully
While the OpenSea sales-feed activity itself cannot work without support from the OpenSea API (by definition), the commands for art-queries (e.g. `#?` or `#100 fidenza`) should be workable in a degraded mode when the OpenSea API goes down.

When an OpenSea API request fails in this case, we should still have artbot respond based on the information that it can get from the Art Blocks APIs, and just omit the OpenSea-specific info

Asupkay:
This change allows for project queries to continue to function in a degraded mode when the OpenSea API is down by falling back to only including nformation available from the AB API. I also moved fields that are available from both APIs to the Art Blocks API exclusively (ex: title) as we have more control over that API being available.

Closes #190


Classes/ProjectBot.js
```diff
@@ -74,23 +74,25 @@ class ProjectBot {
         this.sendMetaDataMessage(openSeaData, msg, tokenID, detailsRequested);
       })
       .catch((err) => {
-        console.error(err);
```
Jbern16, line 77: lets still log a warn that this message was sent in a degraded manner

Jbern16 (approved):
LGTM just minor change to still include a log

Classes/ProjectBot.js
Asupkay: Sounds good!

PR #203, opened 2022-01-31
Fix token features

Asupkay:
Token features were failing as they are no longer an array of strings. This pulls the features from the artBlocksData and if they exist turns them into a key value pair string.


ab-hxm (approved):
lgtm!

PR #211, opened 2022-02-02
Add names to some screens pieces

thomasp85:
This PR adds a few singles names to test out the feature


Asupkay (approved):

jakerockland:
I don't think this will work without also updating your projectBots.json entry to refer to this new screensSingles.json

PR #212, opened 2022-02-02
check of collection_name exists first

Jbern16:
occasionally seeing the following error, which causes a crash. I do believe this is caused when a bundle sells, which is absent of a collection_name.
```
2022-02-02T17:37:15.557053+00:00 app[web.1]:       if (artBlocksData.collection_name.includes('Chromie Squiggle')) {
2022-02-02T17:37:15.557054+00:00 app[web.1]:                                         ^
2022-02-02T17:37:15.557055+00:00 app[web.1]: 
2022-02-02T17:37:15.557055+00:00 app[web.1]: TypeError: Cannot read properties of undefined (reading 'includes')
```

---
  - https://app.asana.com/0/0/1201758951843287

### DIFF
--- a/Utils/activityTriager.js
+++ b/Utils/activityTriager.js
@@ -174,11 +174,11 @@ async function triageActivityMessage(msg, bot) {
     embed.setURL(openseaURL);
 
     // Only forward sales events and listing events.
-    if (eventName.includes('Successful')) {
-      bot.channels.cache.get(CHANNEL_SALES).send(embed);
-      bot.channels.cache.get(CHANNEL_SALES_CHAT).send(embed);
-      if(artBlocksData.collection_name) {
-      // Forward all Chromie Squiggles sales on to the DAO.
+    if  (artBlocksData.collection_name) {
+      if (eventName.includes('Successful')) {
+        bot.channels.cache.get(CHANNEL_SALES).send(embed);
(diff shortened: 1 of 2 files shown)


Utils/activityTriager.js
```diff
@@ -174,30 +173,32 @@ async function triageActivityMessage(msg, bot) {
...
+          bot.channels.cache.get(CHANNEL_SQUIGGLE_SALES).send(embed);
+        }
+        if (artBlocksData.collection_name.includes('Fidenza')) {
+          bot.channels.cache.get(CHANNEL_FIDENZA_AND_IC_SALES).send(embed);
+        }
+        if (artBlocksData.collection_name.includes('Incomplete Control')) {
+          bot.channels.cache.get(CHANNEL_FIDENZA_AND_IC_SALES).send(embed);
+        }
+      } else if (eventName.includes('Created')) {
+        bot.channels.cache.get(CHANNEL_LISTINGS).send(embed);
```
ryley-o, line 191: I think it's okay if this gets skipped for bulk-listings, right?

ryley-o (approved):

Utils/activityTriager.js
Jbern16: they come in as undefined anyway - maybe something to look at as an enhancement?

ab-hxm (commented):
lgtm

PR #215, opened 2022-02-02
Have bot request basic information from subgraph vs. pull from config

### LINKED ISSUE #142: Improve how artbot handles random token selection of partially minted projects
Currently, artbot selects a random `#?` token by using `0` and `Maximum Invocations` as the range for the project.

Ideally, it should use the range of `0` and `Invocations`, which can be known via artblocks subgraph, the artblocks project api, etc. This would prevent selection of non-existent tokens when a user queries `#?` of a project that is only partially minted.

Asupkay:
This PR modifies the artbot to get invocations, contract, and name from the subgraph vs. the config. This allows for the config language to be simplified and queries for a specific token to throw an error instead of returning undefined if that piece has not been minted yet. I do want to eventually merge FactoryBot and ProjectBot together as well as make the config language even simpler and only for naming pieces. That will be in a follow up change as I don't want to make this change too ambiguous.

Closes #142

### DIFF
--- a/ProjectConfig/projectBots.json
+++ b/ProjectConfig/projectBots.json
@@ -215,7 +215,7 @@
     },
     "returnBot": {
         "projectNumber": 77
-    }
+    },
     "ritualBot": {
         "projectNumber": 172
     },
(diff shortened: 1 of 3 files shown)


asaltste (approved):
LGTM :) this is awesome

jakerockland (approved):
This is so rad, absolutely love to see it.

ryley-o:
Awesome!!

PR #218, opened 2022-02-04
[DRAFT] Simplify configuration

Asupkay:
To simplify configuration I switched to using the project id as a key instead of an arbitrary string and instead of creating bots based on the projectBots file I create them by looping over the channels config file and checking what bots the channels need created. This allows us to not need an entry in projectBots if there are no named mappings (or in the future other configuration). This will make it simpler for artists to set up their channels quickly without having to work in two config files. I will be extending the projectBots config file in an upcoming change to allow a contract field. This will be used by PBAB partners to denote that the project id is not from the Art Block contracts.

### DIFF
--- a/ProjectConfig/channels.json
+++ b/ProjectConfig/channels.json
@@ -2,21 +2,21 @@
     "793390526765465600": {
         "name": "hideki-tsukamoto",
         "projectBotHandlers": {
-            "default": "singularityBot"
+            "default": "8"
         }
     },
     "807346227594919986": {
         "name": "ge1doot",
         "projectBotHandlers": {
-            "default": "ignitionBot",
+            "default": "9",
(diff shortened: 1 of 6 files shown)


asaltste (approved):

ryley-o (approved):
Awesome :) I was wondering if this was going to happen :)

can be a separate PR, but we can reflect these recent updates in our ./.github/ISSUE_TEMPLATE/* files as well!

ryley-o:
@Asupkay letting you merge in case you want to watch for anything during deployment

Asupkay:
> can be a separate PR, but we can reflect these recent updates in our ./.github/ISSUE_TEMPLATE/* files as well!

Opened #221. Thanks!!

Issue #169, opened 2021-12-13
Add PBAB support.


jakerockland:

jakerockland:
Note, that this is in large part a documentation issue related to #111 but also more broadly capturing how one invites artbot to their server, etc., as Artbot can and does already live in other servers (e.g. see #173) but does not yet currently have an awareness of PBAB contracts + their API structure.

Issue #108, opened 2021-10-05
Performance improvement - Quick return if message from unrecognized channel


ryley-o:
Create a hashmap of valid Discord channelIDs to respond to, and immediately return if channel will not be responded to. This will result in less checking on every message on channels not being handled by artbot.

Asupkay:
Is this causing the bot to be slow or what is the side effect of handling every message?

ryley-o:
I think the only side effect is more processing than necessary. Jake might have stronger opinions, but seems like a very minor benefit.

jakerockland:
Yeah would just be a minor performance improvement. May be a wontfix for now if we feel like it adds too much unnecessary complexity.

Asupkay:
If the performance improvement is measurable and detectable by the user I'm definitely interested. I just want to make sure this doesn't become more annoying for us where we always have to remember to update the channelID list even though we're not seeing any improvement for the user.

ryley-o:
I think it would be ~no measurable performance improvement for user unless there were so many messages coming in from unwatched channels that artbot started lagging 🤷‍♂️ 

We already specify all watched channels in `/ProjectConfig/projectBots.json`, so shouldn't cause any additional maintenance 😃

Asupkay:
Hmmmm since the watched channels already exist might be worth it if it's super simple. I'll take a look tomorrow and see what I can do. Thanks for the explanation!

Asupkay:
With the expansion to PBAB partners with #230 and #225 this may make a lot of sense. Should be a simple ticket that would be good for someone new.

PR #230, opened 2022-02-09
Add doodle labs project bot config

Asupkay:
This configures art bot to respond to project queries on the Doodle Labs server in the channel called the-lab. This is a walk through of the prototype PBAB instructions for Art Bot to make sure they are solid.

#169


ryley-o (approved):
LGTM!

FWIW - might be a good time to think about an easy fix to #108 if we start adding more servers (although I don't expect it to become a problem)

jakerockland:
Not merging, will let you do so @Asupkay, given that this change requires coordination w/ Doodle Labs team to have artbot invited etc.

Asupkay:
> LGTM!
> 
> FWIW - might be a good time to think about an easy fix to #108 if we start adding more servers (although I don't expect it to become a problem)

Agree that as this expands #108 would be a quick win

PR #234, opened 2022-02-10
Short-circuit message handling if sent in unobserved channel

### LINKED ISSUE #108: Performance improvement - Quick return if message from unrecognized channel

Asupkay:
Note that the current functionality that ignores the channels in channels.json which will be affected by this is smartResponses, and giveways. Giving all the smart responses a once over and thinking about it I don't think there are any egregious ways this restricts where bot messages can be sent. In the worst case we can get more general and allow more channels as time goes on.

Closes #108

### DIFF
--- a/ProjectConfig/projectConfig.js
+++ b/ProjectConfig/projectConfig.js
@@ -203,11 +203,6 @@ class ProjectConfig {
   */
   routeProjectNumberMsg(channelID, msg) {
     const channel = this.channels[channelID];
-    if (!channel) {
-      // only occurs when # messages sent from channels not being observed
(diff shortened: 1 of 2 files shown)


jakerockland (approved):
Rad rad, thank you! :)

asaltste (approved):

ryley-o (approved):
LGTM!

Edit - I should read your PR description - LGTM, maybe Dan or Ponyo have useful thoughts on smartbotresponses

PR #241, opened 2022-02-10
Reimplement "adding config for plottables artist channels"

Asupkay:
An issue arose from the contract address in the config not being all lowercase as when this is passed to the subgraph it was unable to find the projects for Plottables. Created #240 to add a warning to the implementation instructions and considering opening an issue to do one of the following:

A. Refactor the way the bot pulls in addresses to support both versions of the address
B. Fail more gracefully but leave it as invalid
C. Add tests to validate the config is valid before the change can go in

As I see it it's great to keep a consistent format for the addresses; however both forms are technically valid in the ecosystem. To not fully abstract what is going on under the hood and keep it consistent for the PBAB partners working with our subgraph I want to conform to what the subgraph wants. We could fail more gracefully, but waiting all the way till runtime is just bad practice. Ultimately it would be much better to have tests that just validate that the config is valid before runtime and before the PR is merged to prevent this issue and future config issues across the board not just for this one situation. Any feedback is appreciated.

Reverts ArtBlocks/artbot#239


jakerockland (approved):
My personally recommendation in this case would be to file a bug wrt:  C. Add tests to validate the config is valid before the change can go in

PR #252, opened 2022-03-14
Update ProjectBot's mint #0 handling

jakerockland:
Update `ProjectBot` to not return a mint #0 on use of the `#?` command and instead to use mint #1 in the case that mint #0 is randomly selected.

This is done for similar reasons as ArtBlocks/alertbot#60 and https://github.com/ArtBlocks/artbot/pull/251, and continues to build on solving https://app.asana.com/0/1201517173861745/1201967395454694/f

### DIFF
--- a/Classes/ProjectBot.js
+++ b/Classes/ProjectBot.js
@@ -46,6 +46,9 @@ class ProjectBot {
     let pieceNumber;
     if (afterTheHash[0] == "?") {
       pieceNumber = parseInt(Math.random() * this.editionSize);
+      if (pieceNumber === 0) {
+        pieceNumber = 1;
+      }
     } else {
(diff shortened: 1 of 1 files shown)


Jbern16 (approved):

Asupkay (commented):
I feel that in the case where the ProjectBot is triggered by `#?` the use of `#0` is fine as in the situation when the project is not open and this fix is implemented it's going to either:

a. throw an error that the number is invalid as there is only 1 edition minted even though they asked for `#0`
b. get `#1` when the project is not open and have the same error that `#0` has with the link.

Would it make more sense to disallow these types of queries if the project is not open or remove the link completely from the embed for unopen projects? I understand this change for the other places random is used but ProjectBot seems like a special case.

PR #261, opened 2022-05-02
Add LooksRare API Polling

grantoesterling:
## What's New
LooksRare API Polling!
- Artbot will poll LooksRare API every 20 seconds (tentative number, just felt like a happy medium)
- Polls the 25 most recent Sales/Listings for OG/V2 contracts 
-- 4 endpoints
- For each endpoint, it keeps track of the latest event timestamp-wise, and will only report events to Discord if they are newer than the last batch of events it saw

Also added infrastructure to make it easier to add other API Polling bots in the future!

## Screenshots!
Visually the embeds are identical to the current OS ones - just with an added "LooksRare" note in the seller field
Listings:
Screen Shot 2022-05-02 at 5 04 28 PM

Sales:
Screen Shot 2022-05-02 at 5 05 34 PM


grantoesterling:
Apologies for some of these whitespace changes - changes my local Prettier config made that the eslint did not convert back

Classes/ApiPollBot.js
```diff
@@ -0,0 +1,96 @@
...
+   * "Abstact" function each ApiBot must implement - parses endpoint response
+   * @param {*} responseData - Dict parsed from API request json
+   */
+  handleAPIResponse(responseData) {
+    console.warn('handleAPIResponse function not implemented!');
+  }
+}
+
+/** API Poller for LooksRare List and Sale events */
+class LooksRareApiPollBot extends ApiPollBot {
```
jakerockland, line 57: @grantoesterling Thank you very much for making `ApiPollBot` an abstract parent class. 

Could you please move this `LooksRareApiPollBot` implementation to its own implementation file vs. living in the same file as the parent class? Also, nit, can we please use `LooksRareAPIPollBot`.

Classes/ApiPollBot.js
```diff
@@ -0,0 +1,96 @@
+const fetch = require('node-fetch');
+const {triageLooksRareMessage} = require('../Utils/activityTriager');
+
+/** Abstract parent class for all API Poll Bots */
+class ApiPollBot {
```
jakerockland, line 5: Nit: I would prefer `APIPollBot` here.

index.js
```diff
@@ -153,14 +171,49 @@ bot.on('message', (msg) => {
...
+          });
+        }
+      },
+  );
 });
+
+// Instantiate API Pollers
+
+// Rate (in ms) to poll below API endpoints
+const pollTime = 20000;
```
jakerockland, line 191: Nit: prefer we make this `API_POLL_TIME` and define at top of file.

Utils/activityTriager.js
```diff
@@ -175,33 +179,126 @@ async function triageActivityMessage(msg, bot) {
...
+/**
+ * Helper function to send embed message to all proper sales channels
+ * @param {*} bot
+ * @param {*} embed
+ * @param {*} artBlocksData
+ */
+function sendEmbedToSaleChannels(bot, embed, artBlocksData) {
+  bot.channels.cache.get(CHANNEL_SALES).send(embed);
+  bot.channels.cache.get(CHANNEL_SALES_CHAT).send(embed);
+  // // Forward all Chromie Squiggles sales on to the DAO.
```
jakerockland, line 271: Nit: we have an extra `//` here.

Utils/activityTriager.js
```diff
@@ -175,33 +179,126 @@ async function triageActivityMessage(msg, bot) {
...
 
+/**
+ * Parses LooksRare data, builds and sends embed in appropriate channel
+ * API Spec: https://looksrare.github.io/api-docs/#/Events/EventController.getEvents
+ * @param {*} msg - Dict of LooksRare event data
+ * @param {*} bot - Discord bot that will be sending message
+ * @return {void}
+ */
+// eslint-disable-next-line require-jsdoc
+async function triageLooksRareMessage(msg, bot) {
```
jakerockland, line 198: I feel like this functionality should be mostly owned by `LooksRareApiPollBot`, can we refactor to have the `ApiPollBot` conforming-instance be responsible for ineligibly sorting out the incoming message meaning and providing that via a standard interface that can then be triaged via  `sendEmbedToSaleChannels`/`sendEmbedToListChannels`?

jakerockland (changes requested):
This is awesome 🔥 💯 

Have some requested changes below, but this is definitely trending in the right direction and looks great thus far.

Jbern16 (approved):

jakerockland (approved):
Fantastic, thank you very much for the updates @grantoesterling

jakerockland:
LGTM to merge–this is really exciting work.

Reminder that we have to do these deployments manually for now, a'la the convo here: https://discord.com/channels/822311470133542912/930922401808670780/968562463366848513

PR #264, opened 2022-05-04
Adding Test Framework and Config Tests

grantoesterling:
## What's New
- Adding Jest testing framework
- Adding tests for validating config .jsons
- Adding testing as step in CircleCI
- Converting existing FactoryBot tests to Jest
- Updating README

PR Looks big and scary but mostly whitespace changes
Real stuff is in the two test files, plus slightly changing init function of FactoryBot

## Usage
`yarn test`

Please let me know if there are any more config tests to be added!


grantoesterling:
apologies for the million commits - tried to get cute with commit squashing but it was being troublesome so i gave up

jakerockland:
> apologies for the million commits - tried to get cute with commit squashing but it was being troublesome so i gave up

No apology needed–we can squash-on-merge!

.circleci/config.yml
```diff
@@ -18,9 +18,22 @@ jobs:
     executor: build-executor
     steps:
       - checkout
-      - run: |
-          npm install
-          timeout 2m npm start || ( [[ $? -eq 124 ]] && echo "WARNING: Timeout reached, but that's OK" )
```
jakerockland, line 23: @grantoesterling I don't have a strong opinion here–but can you describe why you think we should get rid of this smoke test that just verifies everything runs for a lil?

jakerockland (approved):

jakerockland:
@Asupkay and @Jbern16 were added for visibility, but they don't need to also approve for this to merge IMHO–LGTM

.circleci/config.yml
grantoesterling: eep i meant to leave that in actually - good catch. going to add it back

jakerockland: Ah rad 👍 SGTM

PR #276, opened 2022-05-20
Switching formatter to Prettier!

grantoesterling:
## What's New
- Getting rid of eslint, switching to prettier!

Almost everything in this PR is formatting changes BUT to make the switch I:
- Added prettier to package.json
- Added `yarn format` command
- Added .vscode folder with prettier config stuff
- .prettierrc and .prettierignore files
- Deleted eslinting stuff from README, etc

@Asupkay please lmk if I forgot something!


jakerockland (approved):
Rad :)

Asupkay:
This looks great, and lots of lines changed! One thing I did realize is that we should probably keep both. Eslint is a linter and Prettier is a formatter and although similar, serve distinctly different use cases (more info [here](https://taiyr.me/what-is-the-difference-between-code-linters-and-formatters) if you're interested). In my experience I usually see companies use both at the same time to get all the benefits. They integrate really well and don't conflict so should be super simple. If you look at the artblocks-website repo's [package.json](https://github.com/ArtBlocks/artblocks-website/blob/main/package.json) you can see we use both there and husky to make sure it is run before commit. Can you do a similar setup with husky? At that point both environments should act exactly the same :)

jakerockland:
@grantoesterling when this is merged, could you re-deploy so that we also capture the change merged in #277 in our deployment?

grantoesterling:
@Asupkay Added back eslint and added config to make it play nice with prettier :) merging now!

PR #285, opened 2022-06-09
Adding ABxPace support

grantoesterling:
## What's New
- Adding support for AB x Pace - project query and random query should work (once mint has happened)
- Commonized a bit of GraphQL querying code

## Bench Test
Screen Shot 2022-06-09 at 12 24 54 PM


Utils/parseArtBlocksAPI.js
```diff
@@ -369,15 +369,49 @@ async function getArtBlocksProjects() {
...
+  return await getContractsProjects([PARTNER_CONTRACTS.AB_X_PACE])
+}
+
 /**
  * gets all PBAB Contracts from Hasura
  *
  */
 async function _getPBABContracts() {
+  const nonPBABContracts = Object.values(CORE_CONTRACTS)
+  nonPBABContracts.push(PARTNER_CONTRACTS.AB_X_PACE)
```
Asupkay, line 410: We're going to have more contracts that are collaborations over time and updating this main be a pain. Would it make sense to spit collaborations to their own file so that this keeps itself updated when the config changes.

Asupkay (commented):
Ah see this was already submitted just a thought to clean things up.

Utils/parseArtBlocksAPI.js
grantoesterling: Facts - will open up a mini-PR :)

PR #287, opened 2022-06-13
Adding null check

grantoesterling:
## What's New
- Artbot crashed this morning due to some project it was trying to query having no '.platform' field - just adding a simple null check to prevent this from happening again

Screenshot of the logs from today's crash
Screen Shot 2022-06-13 at 12 06 37 PM


jakerockland (approved):
LGTM–TYVM! 😄 🙏

PR #290, opened 2022-06-20
Archipelago: use reconnecting websocket

decentralion:
WebSocket connections time out after about an hour, so we need to
reconnect


jakerockland (approved):
LGTM! :)

PR #291, opened 2022-06-21
Adding OS APIPoller for Petro

grantoesterling:
## What's New
- OS API Poller for Petro National lists/sales (I already had an OS poller mocked, so this is a quick and dirty solution - will come back to this next week and replace with a more reusable solution)
- Additionally posts sales to the ab-x-pace channel

## Screenshots
Listings
Screen Shot 2022-06-21 at 3 51 20 PM
Sales
Screen Shot 2022-06-21 at 3 51 43 PM


index.js
```diff
@@ -246,3 +247,29 @@ new LooksRareAPIPollBot(
...
+new OpenseaAPIPollBot(
+  `https://api.opensea.io/api/v1/events?collection_slug=${paceSlug}&event_type=created`,
+  API_POLL_TIME_MS,
+  bot,
+  {
+    Accept: 'application/json',
+    'X-API-KEY': process.env.OPENSEA_API_KEY,
+  },
+  COLLAB_CONTRACTS.AB_X_PACE
+)
```
ryley-o, line 275: @grantoesterling you sure we want created?

ryley-o (approved):
LGTM! - one question about if we want to include `created` events (which I think are when new listings are created?)

index.js
ryley-o: (I'm also not up to speed on artbot, so maybe we post created events now? 😆 )

grantoesterling: pompey says yes!

PR #301, opened 2022-07-14
Archipelago bot: add support for ens addresses

decentralion:
When buyers/sellers have an ENS reverse record set, it will now display
their ens name, as in dandelion.eth

### DIFF
--- a/Utils/ens.js
+++ b/Utils/ens.js
@@ -1,0 +1,45 @@
+const namehash = require('@ensdomains/eth-ens-namehash')
+const ethers = require('ethers')
+
+const REGISTRY_ABI = ['function resolver(bytes32 node) view returns (address)']
+const RESOLVER_ABI = [
+  'function addr(bytes32 node) view returns (address)',
+  'function name(bytes32 node) view returns (string)',
+]
+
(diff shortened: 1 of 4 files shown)


Utils/ens.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+async function resolveName(provider, name) {
+  const node = namehash.hash(name)
+  const resolver = await getResolver(provider, node)
+  if (resolver == null) return null
+  return await resolver.addr(node).catch(() => null)
+}
+
+async function lookupAddress(provider, address) {
```
Asupkay, line 28: Instead of contacting the contract directly you should be able to save some code by using the lookupAddress function from the provider see: https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-lookupAddress

Asupkay (changes requested):

Utils/ens.js
Asupkay: Otherwise looks good to me thanks for implementing this!

grantoesterling:
Done here - https://github.com/ArtBlocks/artbot/pull/307

PR #302, opened 2022-07-14
Archipelago bot: don't show PBAB trades

decentralion:
This seems consistent with ArtBot (can't find any Parnassus sales in
block talk, etc)

### DIFF
--- a/Classes/ArchipelagoBot.js
+++ b/Classes/ArchipelagoBot.js
@@ -11,6 +11,8 @@ const {
 
 const WEB_SOCKET_URL = 'wss://api.archipelago.art/ws'
 const COLLECTIONS_API = 'https://api.archipelago.art/v1/market/collections'
+const AB_LEGACY_CONTRACT = '0x059EDD72Cd353dF5106D2B9cC5ab83a52287aC3a'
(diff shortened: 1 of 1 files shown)


Classes/ArchipelagoBot.js
```diff
@@ -11,6 +11,8 @@ const {
 
 const WEB_SOCKET_URL = 'wss://api.archipelago.art/ws'
 const COLLECTIONS_API = 'https://api.archipelago.art/v1/market/collections'
+const AB_LEGACY_CONTRACT = '0x059EDD72Cd353dF5106D2B9cC5ab83a52287aC3a'
```
Asupkay, line 14: Instead of adding the contracts can you import the contract object here?https://github.com/ArtBlocks/artbot/blob/3856691357d3115ac58bfc5bbe8e8481f6fdfef1/ProjectConfig/coreContracts.json

Asupkay (changes requested):

Asupkay (approved):
Besides the nit this looks good to me so I'm going to approve so you don't have to wait for me once you make the change.

Classes/ArchipelagoBot.js
decentralion: Done!

Asupkay (approved):

PR #303, opened 2022-07-14
adding open projects command

grantoesterling:
## What's New
- Adding a command to show random open project :)
- This probably could be done in a cleaner fashion, but i wanted to ship this fast to get on the BT Roulette game lol

### DIFF
--- a/Utils/parseArtBlocksAPI.js
+++ b/Utils/parseArtBlocksAPI.js
@@ -46,6 +46,29 @@ const contractProjects = gql`
   }
 `
 
+const contractOpenProjects = gql`
+  query getContractOpenProjects($id: ID!, $first: Int!, $skip: Int) {
+    contract(id: $id) {
+      projects(
+        first: $first
+        skip: $skip
+        orderBy: projectId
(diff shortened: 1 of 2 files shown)


Utils/parseArtBlocksAPI.js
```diff
@@ -269,6 +292,31 @@ async function _getContractProjects(contractId) {
...
+        .toPromise()
+      allProjects.push(...result.data.contract.projects)
+      if (result.data.contract.projects.length !== maxProjectsPerQuery) {
+        break
+      }
+    }
+    return allProjects
+  } catch (err) {
+    console.error(err)
+    return undefined
```
Asupkay, line 316: nit: you can just do `return` without `undefined` and get the same result

Asupkay (approved):

Utils/parseArtBlocksAPI.js
jakerockland: We should probably have an opinion in a style guide somewhere about this. I have no opinion myself other than an opinion towards standardizing our opinion. cc @yoshiwarab

PR #309, opened 2022-07-26
Switching Pace Pollers to Reservoir

grantoesterling:
## What's new
- Transitioning Opensea Pollers to Reservoir


index.js
```diff
@@ -254,28 +253,25 @@ if (!TEST_MODE) {
...
+      'x-api-key': process.env.RESERVOIR_API_KEY,
     },
     COLLAB_CONTRACTS.AB_X_PACE
   )
 
-  new OpenseaAPIPollBot(
-    `https://api.opensea.io/api/v1/events?collection_slug=${paceSlug}&event_type=created`,
-    API_POLL_TIME_MS,
+  new ReservoirSaleBot(
+    `https://api.reservoir.tools/sales/bulk/v1?contract=${COLLAB_CONTRACTS.AB_X_PACE}&limit=100`,
```
jakerockland, line 269: Could you clarify why limit is 100 here and 50 above? Wonder if we should parameterize this w/ a const.

jakerockland (approved):

index.js
grantoesterling: no reason really - those limits were copied from the example query in Reservoir's docs lol - will parameterize though I think that makes sense

PR #317, opened 2022-08-01
building listing urls

grantoesterling:
## What's New
- Seems like URLs aren't included in Reservoir API responses for listings anymore (at least not consistently), so switching over to what we are doing for the Sale poller (building the URLs, falling back on AB url if unknown marketplace)


Asupkay (approved):

jakerockland:
> Seems like URLs aren't included in Reservoir API responses for listings anymore (at least not consistently), so switching over to what we are doing for the Sale poller

@grantoesterling @Asupkay have we reached out to the Reservoir team about this?

PR #340, opened 2022-08-31
Specifying AB Contracts

grantoesterling:
## What's New
- There was some sort of error happening with the wallet query - I think what was happening was it was also getting PBAB tokens, and the IndexerBot didn't have those projectBots so it errored
- This PR makes it specifically only query for Core Contract tokens
- Also added 10 attempts to get a random project bot just in case (though I don't believe this will need to be used)
- Plus clearer logging
- This can be expanded in the future as we see fit

### DIFF
--- a/Utils/parseArtBlocksAPI.js
+++ b/Utils/parseArtBlocksAPI.js
@@ -142,8 +142,17 @@ const getPBABContracts = gql`
 `
 
 const getWalletTokens = gql`
-  query getWalletTokens($wallet: String!, $first: Int!, $skip: Int) {
-    tokens(first: $first, skip: $skip, where: { owner: $wallet }) {
+  query getWalletTokens(
+    $wallet: String!
+    $contracts: [ID!]!
+    $first: Int!
+    $skip: Int
+  ) {
+    tokens(
+      first: $first
+      skip: $skip
+      where: { owner: $wallet, contract_in: $contracts }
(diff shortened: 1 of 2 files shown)


Utils/parseArtBlocksAPI.js
```diff
@@ -620,6 +629,7 @@ async function getAllWalletTokens(walletAddress) {
       const result = await client
         .query(getWalletTokens, {
           wallet: walletAddress,
+          contracts: Object.values(CORE_CONTRACTS),
```
jakerockland, line 632: I think we probably want this to be the combined set of COLLAB_CONTRACTS, and CORE_CONTRACTS?

jakerockland (approved):
One comment on scope here–otherwise LGTM.

Utils/parseArtBlocksAPI.js
grantoesterling: I do not believe that will work - the collab_contracts are in their own ArtIndexerBot, so that will require a separate fix - i'm gonna look into that right now though

grantoesterling: actually yeah that definitely won't work - that is a bigger issue though, probably for a different PR

jakerockland: SGTM ser.

PR #364, opened 2022-09-23
Adding collection to lists and sales

grantoesterling:
## What's New
- Adding Curation Status to list/sales events

Screen Shot 2022-09-23 at 5 37 28 PM

### DIFF
--- a/Classes/APIBots/ReservoirListBot.js
+++ b/Classes/APIBots/ReservoirListBot.js
@@ -86,8 +86,14 @@ class ReservoirListBot extends APIPollBot {
     const artBlocksResponse = await fetch(tokenUrl)
     const artBlocksData = await artBlocksResponse.json()
 
+    const curationStatus = artBlocksData?.curation_status
+      ? artBlocksData.curation_status[0].toUpperCase() +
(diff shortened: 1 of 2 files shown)


Classes/APIBots/ReservoirListBot.js
```diff
@@ -86,8 +86,14 @@ class ReservoirListBot extends APIPollBot {
     const artBlocksResponse = await fetch(tokenUrl)
     const artBlocksData = await artBlocksResponse.json()
 
+    const curationStatus = artBlocksData?.curation_status
+      ? artBlocksData.curation_status[0].toUpperCase() +
```
Jbern16, line 90: should be using vertical_name here instead

Jbern16 (changes requested):

mchrupcala (approved):

Classes/APIBots/ReservoirListBot.js
grantoesterling: @Jbern16 the token API doesn't have `vertical_name`, it has `curation_status`

Jbern16 (approved):

PR #385, opened 2022-10-10
fixing pace sales

grantoesterling:
HOTFIX
- Setting collection name for pace sales/lists to stop Artbot from crashing


Jbern16 (approved):

jakerockland (approved):
Fine for this hotfix, but these should be shared constants somewhere I think, even just within the scope of this repo. Could you file a bug for future follow up please?

jakerockland:
Thank you for the quick fix 💜 🙌

PR #391, opened 2022-10-20
Mint Bot!!

grantoesterling:
## What's New
- Adding ability for artbot to report mints (and eventually replace alertbot)
- The structure is extremely similar to alertbot:
    - Added `/new-mint` POST endpoint to Artbot
    - After some initial checks to make sure everything is in order, it adds it to a Set in the new `MintBot` class
     - Every X seconds, it'll go through any new messages in that queue, hit the Token API to get the image+generator endpoints, then hit the image endpoint to ensure the image is actually valid
     - If image is not ready, that mint stays in the queue
     - If the image is ready, mint is removed from queue and discord message is sent to proper channels!
   
## TODO Post Merge
- Switch mainnet Hasura env var ALERTBOT_WEBHOOK to be new value!
- Take alertbot offline 🔫 

Screen Shot 2022-10-31 at 11 12 49 AM


Jbern16:
@grantoesterling how does this effect Alertbot as a service for engine partners?

grantoesterling:
> @grantoesterling how does this effect Alertbot as a service for engine partners?

@Jbern16 great question - i believe the partners just need to 1. Add artbot to their server, 2. Add their mint channel to channels.json, 3. add an entry in the mintBotConfig pointing their contract to their channel, and it should all work. I'll add those as part of this PR, but should definitely contact Plottables and whoever else to make them aware of this

Jbern16:
> Every X seconds, it'll go through that queue and hit the token API to see if the image has been rendered yet (may consider replacing this with a hasura query)

I think we should probably use the hasura webhook that reports when image id is ready instead. We used to hit the token api, and decided to refactor this logic. Long render times on projects makes this a a relatively inefficient way to get this data consistently.

Jbern16 (approved):

PR #394, opened 2022-10-24
adding smartbot response for verticals

grantoesterling:
## What's New
- Modifying response for when people ask about the verticals
- Copy from Massoud

Screen Shot 2022-10-24 at 4 08 09 PM

### DIFF
--- a/Utils/smartBotResponse.js
+++ b/Utils/smartBotResponse.js
@@ -128,24 +128,31 @@ const OPENSEA_CURATED_MESSAGE = new MessageEmbed()
   )
 
 /*
- * Custom message shown when someone asks what the "Playground" vs. "Curated"
- * vs. "Factory" is.
+ * Custom message shown when someone asks what the "Heritage" vs. "Curated"
+ * vs. "Explorations", etc is.
  */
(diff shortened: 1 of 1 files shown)


Utils/smartBotResponse.js
```diff
@@ -151,8 +151,17 @@ const VERTICALS_MESSAGE = new MessageEmbed()
     **Art Blocks x Collaborations** are with select partners and organized into a new vertical of co-branded collections. Because partner projects live directly in the Art Blocks ecosystem, they will be subject to the same quality review and scheduling as any other Art Blocks project. These collections represent true collaborations between Art Blocks and our partners to shape the program of offerings.
     
     **Art Blocks Heritage** is a designation to denote Art Blocks projects released as either Playground or Factory projects before 10/4/22.
```
jakerockland, line 153: I don't think this is right–should be before 2022-11-01. cc @sarahrossien 

Can we fix this with this PR please?

jakerockland (approved):

mchrupcala (approved):
Looks good! Approved pending Jake's comment re: the date.

PR #398, opened 2022-10-25
New #? commands!! (PLS MERGE BY 10/26)

grantoesterling:
## What's New
- Added `#?` commands for the new verticals:
    - `heritage`, `presents`, `collaborations/collabs`, `explorations` (won't start working til it is active and more tokens are minted don't worry), `factory`, `playground`, `curated series x`
    - All of these should also work with wallets
- Also added the new core contracts

### DIFF
--- a/Classes/APIBots/utils.js
+++ b/Classes/APIBots/utils.js
@@ -103,17 +103,18 @@ function isVerticalName(msg) {
     msg === 'presents' ||
     msg === 'collaborations' ||
     msg === 'collabs' ||
-    msg === 'heritage'
+    msg === 'heritage' ||
+    msg === 'factory' ||
+    msg === 'playground' ||
+    msg === 'explorations' ||
+    msg === 'engine' ||
+    msg.startsWith('curatedseries')
   )
 }
 function getVerticalName(msg) {
   switch (msg) {
     case 'collabs':
       return 'collaborations'
-    case 'collaborations':
-    case 'heritage':
-    case 'curated':
-    case 'presents':
     default:
       return msg
   }
--- a/Classes/APIBots/utils.js
+++ b/Classes/APIBots/utils.js
@@ -76,7 +76,25 @@ function isWallet(msg) {
 }
 
 function isVerticalName(msg) {
-  return msg === 'curated' || msg === 'factory' || msg === 'playground'
+  return (
+    msg === 'curated' ||
+    msg === 'presents' ||
+    msg === 'collaborations' ||
+    msg === 'collabs' ||
+    msg === 'heritage'
+  )
+}
+function getVerticalName(msg) {
+  switch (msg) {
+    case 'collabs':
+      return 'collaborations'
+    case 'collaborations':
+    case 'heritage':
+    case 'curated':
+    case 'presents':
+    default:
+      return msg
+  }
 }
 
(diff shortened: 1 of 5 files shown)


Classes/APIBots/utils.js
```diff
@@ -98,11 +98,31 @@ function isWallet(msg) {
...
+    msg === 'curated' ||
+    msg === 'presents' ||
+    msg === 'collaborations' ||
+    msg === 'collabs' ||
+    msg === 'heritage' ||
+    msg === 'factory' ||
+    msg === 'playground' ||
+    msg === 'explorations' ||
+    msg === 'engine' ||
+    msg.startsWith('curatedseries')
```
Jbern16, line 111:
```suggestion
    msg.startsWith('curated series')
```

I think there should be a space?

Utils/parseArtBlocksAPI.js
```diff
@@ -596,10 +605,18 @@ async function getProjectsCurationStatus() {
       }
     }
     const curationMapping = {}
+    const heritageStatuses = {}
     allProjects.forEach((proj) => {
-      curationMapping[proj.id] = proj.vertical_name.toLowerCase()
+      curationStatus = proj.vertical_name.toLowerCase()
+      if (proj.vertical?.category_name?.toLowerCase() !== 'collections') {
+        curationStatus = proj.vertical?.category_name?.toLowerCase()
```
Jbern16, line 612: hm, whats the logic here? Curation Status should always reflect vertical/name, never vertical/category_name

Jbern16 (approved):
had a few questions but lgtm

Classes/APIBots/utils.js
grantoesterling: nope, since artbot has already deburred and de-whitespaced the message by this point

Utils/parseArtBlocksAPI.js
grantoesterling: It's basically a hack to have all abxpace and fullyonchain projects be reporting under `#? collaborations` and `#? engine` respectively

This artbot feature is a bit misleading, as some of these commands are for Categories, and some are for Vertical Names / Curation Status

basically `curationStatus` is the wrong variable name to use here lol, should probs rename to `collection`

jakerockland: ++; seems worth the rename

Classes/APIBots/utils.js
```diff
@@ -98,11 +98,31 @@ function isWallet(msg) {
...
+  return (
+    msg === 'curated' ||
+    msg === 'presents' ||
+    msg === 'collaborations' ||
+    msg === 'collabs' ||
+    msg === 'heritage' ||
+    msg === 'factory' ||
+    msg === 'playground' ||
+    msg === 'explorations' ||
+    msg === 'engine' ||
```
jakerockland, line 110: Want to confirm, does this also capture Engine Flex?

jakerockland (approved):

Classes/APIBots/utils.js
grantoesterling: yes it will (engine command won't be available in BT though til a later PR refactoring the initialization of this stuff)

When I add that engine support to BT, I'll also add the `fullyonchain` and `flex` commands too, but that was lower priority so decided to not include it here

PR #405, opened 2022-10-30
Allow block-talkers to thank Grant in Spanish and German!

pmohun:
Grant deserves to be thanked in many languages, not just English


mchrupcala (approved):
Love it!

PR #408, opened 2022-11-01
changing sales to use fill source

grantoesterling:
Changing Reservoir sale post to use fill source instead of order source (so Sansa and other such marketplaces will be respected!


jakerockland (approved):
!vibe

PR #409, opened 2022-11-01
remove explorations from core contracts

Jbern16:


jakerockland (approved):
LGTM pending confirmation / resolution. Of the thing I flagged in Slakc

PR #422, opened 2022-11-08
Adding Engine mints and sales

grantoesterling:
Adding Engine mints and sales (to respective channels)

### DIFF
--- a/src/Classes/APIBots/ReservoirSaleBot.ts
+++ b/src/Classes/APIBots/ReservoirSaleBot.ts
@@ -1,5 +1,9 @@
 import { Client } from 'discord.js'
-import { getTokenApiUrl, isExplorationsContract } from './utils'
+import {
+  getTokenApiUrl,
+  isExplorationsContract,
+  isEngineContract,
+} from './utils'
 
 const { APIPollBot } = require('./ApiPollBot')
 const { EmbedBuilder } = require('discord.js')
@@ -106,7 +110,6 @@ export class ReservoirSaleBot extends APIPollBot {
     if (sale.orderKind === 'mint') {
       return // Don't send mint events
     }
-
     const priceText = 'Sale Price'
     const price = sale.price.amount.decimal
     const currency = sale.price.currency.symbol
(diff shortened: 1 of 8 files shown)


src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -130,7 +133,7 @@ export class ReservoirSaleBot extends APIPollBot {
 
     let sellerText = await this.ensOrAddress(sale.from)
     let buyerText = await this.ensOrAddress(sale.to)
-    let platformUrl = artBlocksData.external_url
+    let platformUrl = artBlocksData.external_url ?? 'https://www.artblocks.io'
```
jakerockland, line 136: QQ: will this cause Engine tokens to be posted with flagship links? If so, that isn't expected to be a pathway that is available.

jakerockland (approved):

src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -182,6 +188,11 @@ export class ReservoirSaleBot extends APIPollBot {
       curationStatus = 'AB x Pace'
     } else if (isExplorationsContract(sale.token.contract)) {
       curationStatus = 'Explorations'
+    } else if (await isEngineContract(sale.token.contract)) {
```
ryley-o, line 191: Does this cause us to ping our api to get all contracts during every reservoir sale? maybe we could implement a simple caching layer if this causes us to add a lot of requests?

ryley-o (approved):

src/Classes/APIBots/ReservoirSaleBot.ts
grantoesterling: only if the external_url field is empty in the token api response (which indeed it is for some partners, like doodle labs)

I don't really see a feasible recourse if this data isn't in the token api, other than having a hardcoded dict

src/Classes/APIBots/ReservoirSaleBot.ts
grantoesterling: nope! I'm getting every engine contract just once in the `index.ts`, this just needs to be an await since technically that first call is a promise

so only one API ping!

src/Classes/APIBots/ReservoirSaleBot.ts
grantoesterling: i like that solution!

PR #432, opened 2022-11-29
Adding FB batch messages

grantoesterling:
Adding batching to FB sales (if there are many sales by a single wallet in one poll batch, combine them into one message)

Screenshot 2022-11-29 at 4 42 09 PM

### DIFF
--- a/src/Classes/APIBots/ReservoirSaleBot.ts
+++ b/src/Classes/APIBots/ReservoirSaleBot.ts
@@ -4,6 +4,7 @@ import {
   BAN_ADDRESSES,
   sendEmbedToSaleChannels,
 } from '../../Utils/activityTriager'
+import { CollectionType } from '../MintBot'
 import { APIPollBot } from './ApiPollBot'
 import {
   getTokenApiUrl,
@@ -29,6 +30,7 @@ type ReservoirSale = {
     }
     amount: {
       decimal: number
+      native: number
     }
   }
   timestamp: number
@@ -71,14 +73,22 @@ export class ReservoirSaleBot extends APIPollBot {
    */
   async handleAPIResponse(responseData: ReservoirSaleResponse) {
     let maxTime = 0
+    const sales: { [id: string]: ReservoirSale[] } = {}
     for (const data of responseData.sales) {
       const eventTime = data.timestamp
       // Only deal with event if it is new and unique saleId
       if (this.lastUpdatedTime < eventTime && !this.saleIds.has(data.saleId)) {
-        this.buildDiscordMessage(data).catch((err) => {
-          console.log('Error sending sale message', err)
-        })
         this.saleIds.add(data.saleId)
+        if (!isExplorationsContract(data.token.contract)) {
+          this.buildDiscordMessage(data).catch((err) => {
+            console.log('Error sending sale message', err)
+          })
+        } else {
+          if (!sales[data.to]) {
+            sales[data.to] = []
+          }
+          sales[data.to].push(data)
+        }
       }
 
       // Save the time of the latest event from this batch
@@ -86,6 +96,18 @@ export class ReservoirSaleBot extends APIPollBot {
         maxTime = eventTime
       }
     }
+    // Handle Explorations sales (may need to batch big sweeps)
+    Object.keys(sales).forEach((user) => {
+      if (sales[user].length > 1) {
+        this.buildSweepDiscordMessage(sales[user]).catch((err) => {
+          console.log('Error sending batch sale message', err)
+        })
+      } else {
+        this.buildDiscordMessage(sales[user][0]).catch((err) => {
+          console.log('Error sending sale message', err)
+        })
+      }
+    })
 
     // Update latest time vars if batch has new latest time
     if (maxTime > this.lastUpdatedTime) {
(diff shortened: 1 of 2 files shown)


src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -71,21 +73,41 @@ export class ReservoirSaleBot extends APIPollBot {
...
-        })
         this.saleIds.add(data.saleId)
+        if (!isExplorationsContract(data.token.contract)) {
+          this.buildDiscordMessage(data).catch((err) => {
+            console.log('Error sending sale message', err)
+          })
+        } else {
+          if (!sales[data.to]) {
+            sales[data.to] = []
+          }
```
jakerockland, line 89: This reads odd enough that its worth adding a comment that clarifies that this is here to ensure that a null object isn't passed in.

src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -71,21 +73,41 @@ export class ReservoirSaleBot extends APIPollBot {
...
+    const sales: { [id: string]: ReservoirSale[] } = {}
     for (const data of responseData.sales) {
       const eventTime = data.timestamp
       // Only deal with event if it is new and unique saleId
       if (this.lastUpdatedTime < eventTime && !this.saleIds.has(data.saleId)) {
-        this.buildDiscordMessage(data).catch((err) => {
-          console.log('Error sending sale message', err)
-        })
         this.saleIds.add(data.saleId)
+        if (!isExplorationsContract(data.token.contract)) {
```
jakerockland, line 82: Worth adding a dev comment here that clarifies why we skip here.

src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -227,4 +249,82 @@ export class ReservoirSaleBot extends APIPollBot {
...
+          sale.token.tokenId
+        )
+      } else if (platform.includes('looksrare')) {
+        platformUrl = this.buildLooksRareURL(
+          sale.token.contract,
+          sale.token.tokenId
+        )
+      } else if (platform.includes('x2y2')) {
+        platformUrl = this.buildX2Y2URL(sale.token.contract, sale.token.tokenId)
+      }
```
jakerockland, line 304: Can we implement this in a way that is more DRY with regards to https://github.com/ArtBlocks/artbot/blob/b84106ace826faedf990c07fb5e282fae29f8dd3/src/Classes/APIBots/ReservoirSaleBot.ts#L161-L183 please ser? 🙏

jakerockland (approved):
Some minor requests re: comment nits and requested refactor – otherwise LGTM! Thank you @grantoesterling !!

src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -227,4 +249,82 @@ export class ReservoirSaleBot extends APIPollBot {
       )
     }
   }
+
+  async buildSweepDiscordMessage(sales: ReservoirSale[]) {
+    if (BAN_ADDRESSES.has(sales[0].to)) {
```
lindsgil, line 254: nit: we use `sales[0]` so much in this func that it's prob worth creating a var at the start, `const currSale = sales[0]`

src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -227,4 +249,82 @@ export class ReservoirSaleBot extends APIPollBot {
...
+
+    let totalCost = 0
+
+    for (let i = 0; i < sales.length; i++) {
+      const sale = sales[i]
+      if (sale.orderKind === 'mint') {
+        return // Don't send mint events
+      }
+      const tokenName =
+        'Friendship Bracelets #' + (parseInt(sale.token.tokenId) % 1e6)
```
lindsgil, line 280: can we get the token name dynamically instead of hard-coding it here?

grantoesterling: we actually can't unless we hit the token api for every token which i was trying to avoid :/ Reservoir doesn't have that info in the sale data unfortunately

lindsgil (approved):

src/Classes/APIBots/ReservoirSaleBot.ts
```diff
@@ -71,21 +73,46 @@ export class ReservoirSaleBot extends APIPollBot {
...
+    const sales: { [id: string]: ReservoirSale[] } = {}
     for (const data of responseData.sales) {
       const eventTime = data.timestamp
       // Only deal with event if it is new and unique saleId
       if (this.lastUpdatedTime < eventTime && !this.saleIds.has(data.saleId)) {
-        this.buildDiscordMessage(data).catch((err) => {
-          console.log('Error sending sale message', err)
-        })
         this.saleIds.add(data.saleId)
+        // Only worrying about batch sales messages for Friendship Bracelets
```
Asupkay, line 82: Any reason to do this just on friendship bracelets and not all collections? It seems like the problem is that there are too many sales at once that are clogging the chat which isn't something that's been isolated to friendship bracelets historically. Back in August last year people wanted factory to be removed from block-talk so we'll probably see this manifest in other collections as well.

grantoesterling: This is just a band-aid solution to see if we can fix the FB deluge. i wanted to get this out to see if it was even effective / worth it at all before spending more time on it. Also currently this problem only exists with FB - I'd argue sweeps of other collections should be posted as normal as it is more exciting and rare

The good news is it will be relatively easy to add other projects to this in the future

PR #434, opened 2022-12-01
Adding error handling for rate limit

grantoesterling:
Adding error handling for Token API rate limits on `#?` commands

### DIFF
--- a/src/Classes/ProjectBot.ts
+++ b/src/Classes/ProjectBot.ts
@@ -1,3 +1,4 @@
+import { AxiosError } from 'axios'
 import { Message } from 'discord.js'
 
 const { EmbedBuilder } = require('discord.js')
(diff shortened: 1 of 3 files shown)


src/Classes/ProjectBot.ts
```diff
@@ -113,7 +114,20 @@ export class ProjectBot {
 
     const tokenID = pieceNumber + this.projectNumber * 1e6
 
-    this.sendMetaDataMessage(msg, tokenID.toString(), detailsRequested)
+    this.sendMetaDataMessage(msg, tokenID.toString(), detailsRequested).catch(
+      (err: Error | AxiosError) => {
+        console.log('Error sending listing message', err)
+        if (axios.isAxiosError(err)) {
+          const axErr = err as AxiosError
+          // If Token API is being rate limited, send apology note
```
Asupkay, line 122: nit: comment isn't necessary as the code below is self descriptive with the error message. I try to save comments for situations in which code is confusing otherwise people get comment fatigue and stop reading them or maintaining them. This leads to outdated comments across the codebase and makes them less effective.

src/Classes/ProjectBot.ts
```diff
@@ -113,7 +114,20 @@ export class ProjectBot {
 
     const tokenID = pieceNumber + this.projectNumber * 1e6
 
-    this.sendMetaDataMessage(msg, tokenID.toString(), detailsRequested)
+    this.sendMetaDataMessage(msg, tokenID.toString(), detailsRequested).catch(
+      (err: Error | AxiosError) => {
+        console.log('Error sending listing message', err)
```
Asupkay: Should be console.error instead of console.log

Asupkay (approved):

src/Classes/ProjectBot.ts
```diff
@@ -113,7 +114,20 @@ export class ProjectBot {
 
     const tokenID = pieceNumber + this.projectNumber * 1e6
 
-    this.sendMetaDataMessage(msg, tokenID.toString(), detailsRequested)
+    this.sendMetaDataMessage(msg, tokenID.toString(), detailsRequested).catch(
```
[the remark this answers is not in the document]
grantoesterling: i have seen weird behavior with this before when developing (albeit, that might have been when this was still in js) that leads me to want to put the catch here - something about odd async behavior and not awaiting would slip the try catch in weird situations

PR #449, opened 2023-01-24
updating readme

grantoesterling:
Updating readme with GitPOAP and changes from the past few months

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ArtBot: The Art Blocks Discord Bot
 
-[![CircleCI](https://circleci.com/gh/ArtBlocks/artbot/tree/main.svg?style=svg)](https://circleci.com/gh/ArtBlocks/artbot/tree/main)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # ArtBot: The Art Blocks Discord Bot
 
-[![CircleCI](https://circleci.com/gh/ArtBlocks/artbot/tree/main.svg?style=svg)](https://circleci.com/gh/ArtBlocks/artbot/tree/main)
+[![GitPOAPs](https://public-api.gitpoap.io/v1/repo/ArtBlocks/artbot/badge)](https://www.gitpoap.io/gh/ArtBlocks/artbot)
```
jakerockland, line 3: Why don't we include both of these badges?

jakerockland (approved):
LGTM, with the one note that removing the Circle badge seems wrong – can't we just have both?

README.md
grantoesterling: Because we don't use CircleCI for CI anymore, we use Github Actions now. Good point tho I'll add in a badge for Github Actions!!

jakerockland: Ah – roger that. Good call though ser yes pls.

PR #456, opened 2023-02-03
Converted smartBotResponse.js to typescript

izzy1090:
Currently left smartBotResponse.js in the Utils folder so you can refer back to it if needed. Happy to remove the file and commit again if necessary.

Overall types were added on variables and functions where it's clear what the intended type should be. Finally I exported the smartBotResponse func and changed the 'dotenv' import to be consistent with other TS files.


grantoesterling (commented):
Looks good! Can you remove the .js file as part of this PR though?

izzy1090:
> Looks good! Can you remove the .js file as part of this PR though?

Done!

grantoesterling (approved):

src/Utils/smartBotResponse.ts
```diff
@@ -0,0 +1,482 @@
...
+ * @param {*} msgAuthor
+ * @param {*} artBotID
+ * @param {*} channelID
+ */
+export async function smartBotResponse(
+  msgContentLowercase: string,
+  msgAuthor: string,
+  artBotID: string,
+  channelID: string
+) {
```
mchrupcala, line 269: I feel like it could be really nice to write a custom SmartBotReturn type object and define the potential return types in here (string for 'gm', EmbedBuilder, etc) but that would be for a separate anyway PR since we're just doing the JS-TS conversion here. @grantoesterling what do you think?

mchrupcala (approved):

src/Utils/smartBotResponse.ts
grantoesterling: trueeee I didn't think about that - could just do like `string | EmbedBuilder` though, idk a reason we'd need a whole struct thingy

mchrupcala: Ah yeah very true, that works! I just think adding type-safety to the file's res could help us catch anything weird down the road.

PR #462, opened 2023-02-08
Adding Twitter posting to artbot

grantoesterling:
Final step in the alertbot deprecation!! Adding Twitter posting to artbot
- Code taken from alertbot repo and cleaned up
- Tested on dev and everything works!

Still TODO:
Handle Engine project posting - but I think we can cross that bridge when we get there?

### DIFF
--- a/src/Classes/MintBot.ts
+++ b/src/Classes/MintBot.ts
@@ -3,6 +3,7 @@ import { ENGINE_CONTRACTS, mintBot } from '../index'
 import axios, { AxiosError } from 'axios'
 import { getTokenApiUrl } from './APIBots/utils'
 import { ensOrAddress } from './APIBots/utils'
+import { TwitterBot } from './TwitterBot'
 
 const projectConfig = require('../ProjectConfig/projectConfig').projectConfig
 const MINT_CONFIG: {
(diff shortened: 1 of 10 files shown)


src/Classes/MintBot.ts
```diff
@@ -27,13 +28,18 @@ export enum CollectionType {
...
+  twitterBot?: TwitterBot
   newMints: { [id: string]: Mint } = {}
   mintsToPost: { [id: string]: Mint } = {}
   contractToChannel: { [id: string]: string[] } = {}
   constructor(bot: Client) {
     this.bot = bot
     this.buildContractToChannel()
     this.startRoutine()
+
+    if (!process.env.TEST_MODE && process.env.AB_TWITTER_API_KEY) {
```
jakerockland, line 40: I think we should use `process.env.PRODUCTION_MODE` instead so that the default value here isn't falsey in the wrong direction, avoiding us accidentally shipping this live if we forgot the flag (default behavior if we forget the mode flag would be off).

jakerockland (approved):

Asupkay (approved):

src/Classes/MintBot.ts
grantoesterling: I like that

PR #466, opened 2023-02-22
Adding BM collab contract

grantoesterling:
Adding BM collab contract and adjusting existing collab logic


src/Classes/APIBots/ReservoirListBot.ts
```diff
@@ -146,6 +146,8 @@ export class ReservoirListBot extends APIPollBot {
 
     if (artBlocksData?.platform === 'Art Blocks x Pace') {
       curationStatus = 'AB x Pace'
+    } else if (artBlocksData?.platform === 'Art Blocks × Bright Moments') {
```
Jbern16, line 149: can we make a followup to standardize this for all platforms (lower case + trim spaces)

Jbern16 (approved):

src/Classes/APIBots/ReservoirListBot.ts
grantoesterling: Errr what exactly do you mean? Not sure how we'd standardize other than replacing "Art Blocks" with AB

PR #467, opened 2023-02-22
Adding decentralized subgraph option

grantoesterling:
Adding decentralized subgraph as option (toggled by env var)

env vars have been added into heroku already


Asupkay (approved):

mchrupcala (approved):
Looks great! I wonder if this could alternatively be just one API_URL env var that we can switch (locally? in Github?) based on where we'd like to point.

PR #492, opened 2023-04-10
MP4 Embeds To Default to GIF

mchrupcala:
Projects that support MP4 previews should default to show GIF in artbot so we don't overdo it with Discord.

---
  - https://app.asana.com/0/0/1204366239491936

### DIFF
--- a/src/Classes/APIBots/ReservoirListBot.ts
+++ b/src/Classes/APIBots/ReservoirListBot.ts
@@ -161,7 +161,7 @@ export class ReservoirListBot extends APIPollBot {
     }
     // Update thumbnail image to use larger variant from Art Blocks API.
     if (artBlocksData?.image && !artBlocksData.image.includes('undefined')) {
-      replaceVideoWithGIF(artBlocksData.image)
+      artBlocksData.image = replaceVideoWithGIF(artBlocksData.image)
       embed.setThumbnail(artBlocksData.image)
     }
 
--- a/src/Classes/APIBots/ReservoirListBot.ts
+++ b/src/Classes/APIBots/ReservoirListBot.ts
@@ -13,6 +13,7 @@ import {
   getTokenApiUrl,
   isEngineContract,
   isExplorationsContract,
+  replaceVideoWithGIF,
 } from './utils'
 
 type ReservoirListing = {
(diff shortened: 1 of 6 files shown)


src/Classes/ProjectBot.ts
```diff
@@ -197,6 +197,10 @@ export class ProjectBot {
       ownerText = ownerText.substring(0, 6) + '...' + ownerText.substring(38)
     }
 
+    if (artBlocksData.image.includes('mp4')) {
+      artBlocksData.image.replace('mp4', 'gif')
```
lindsgil, line 201: nit: can we share this fn and throw a comment on top of it, something like "discord embeds don't support mp4 rn so we're always sending GIFs"

lindsgil (approved):

src/Classes/ProjectBot.ts
grantoesterling: ++

grantoesterling: you'll need to do this in the sales/list/mint bots as well - so probs throw this in the utils file

mchrupcala: Good call! I made those changes. @grantoesterling we'll always see a delay between the image renders and video renders (since vids go to the low-priority queue)...so maybe we shouldn't add this in MintBot? And we use `image` in TwitterBot too but it looks like that account's down right now.

grantoesterling: @mchrupcala I think it makes sense to make the changes anywhere that image field is being used, even if it is very unlikely a video appears (like in MintBot)

Jbern16 (approved):

src/Classes/APIBots/ReservoirListBot.ts
```diff
@@ -160,6 +161,7 @@ export class ReservoirListBot extends APIPollBot {
     }
     // Update thumbnail image to use larger variant from Art Blocks API.
     if (artBlocksData?.image && !artBlocksData.image.includes('undefined')) {
+      replaceVideoWithGIF(artBlocksData.image)
```
grantoesterling, line 164: need to be setting the variable here

grantoesterling (approved):

src/Classes/ProjectBot.ts
mchrupcala: Grant & I chatted and TL;DR we're handling this in MintBot, but we're just leaving a comment for now in TwitterBot because there's twitter-specific file encoding for the uploads and bc that logic has a bug that needs some extra work in order to fix 🚀 🚀

PR #503, opened 2023-04-27
[DRAFT] Add basic art GPT

jakerockland:
This is an experiment and very much a WIP.


jakerockland:
Leaving this in draft mode for now, as there are some minor param/config optimizations/modifications that @lyaunzbe and I are still adjusting.

However, I think the fundamental structure and architecture here is to a point where we can get an initial review pass from you @grantoesterling when you have the bandwidth to do so 🙏 

Initial thinking is rolling this out on a pretty limited basis (may just be in the internal sever and in a private channel for internal purposes), so would encourage a review with that context in mind.

src/Classes/ArtGPTBot.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Helper to determine if the bot is being queries in valid server and channel.
+   */
+  inValidServerChannel(msg: Message): boolean {
+    const serverID = msg.guild ? msg.guild.id : ''
+    const channelID = msg.channel ? msg.channel.id : ''
+
+    // For now, we only handle messages in the Inc and test servers
+    if (serverID == DISCORD_TEST_SERVER_ID) {
```
grantoesterling, line 140: I believe these should all be `===` instead of `==`

src/Classes/ArtGPTBot.ts
```diff
@@ -0,0 +1,306 @@
...
+// NOTE: OPENAI_API_KEY is not needed to be imported directly,
+//       but it is assumed by Langchain to be available in `process.env`
+//       so must be present in the .`env` file.
+
+// ArtBot username
+const ARTBOT_USERNAME = 'artbot'
+const ARTBOT_MAX_CHARS_RESPONSE = 4000
+
+// Discord consts
+const DISCORD_TEST_SERVER_ID = '785144843986665472'
```
grantoesterling, line 23: I imagine this will be included in the remaining config changes you mentioned, but these should be pulled from the `channels.json` - perhaps we can add an optional field for channels where we want this enabled?

src/Classes/ArtGPTBot.ts
```diff
@@ -0,0 +1,306 @@
...
+    const serverID = msg.guild ? msg.guild.id : ''
+    const channelID = msg.channel ? msg.channel.id : ''
+
+    // For now, we only handle messages in the Inc and test servers
+    if (serverID == DISCORD_TEST_SERVER_ID) {
+      // Handle all messages in the test server
+      return true
+    } else if (serverID == DISCORD_INC_SERVER_ID) {
+      // Only handle messages in the specified #artgpt channel
+      if (channelID == DISCORD_INC_ARTGPT_CHANNEL_ID) {
```
grantoesterling, line 145: Note: this channel still needs to be in channels.json, or else the code will never reach this point, and will be filtered out by code in `index.ts`
Screenshot 2023-05-02 at 4 16 23 PM

src/Classes/ArtGPTBot.ts
```diff
@@ -0,0 +1,306 @@
...
+  /**
+   * Send an embed reply to a message.
+   * @param msg The message to reply to.
+   */
+  async handleRequest(msg: Message) {
+    /*
+     * NOTE: It is important to check if the message author is the ArtBot
+     *       Itself to avoid a recursive infinite loop.
+     */
+    if (msg.author.username == ARTBOT_USERNAME) {
```
grantoesterling, line 161: nit: ===

src/Classes/ArtGPTBot.ts
```diff
@@ -0,0 +1,306 @@
...
+      sourceLocations.forEach((location: any) => {
+        sourceLocationsString += `
+        - ${location.fileName} in ${location.repoName}
+        `
+      })
+
+      // Summarize response to be less than ARTBOT_MAX_CHARS_RESPONSE if it is too long.
+      if (
+        response.text.length + sourceLocationsString.length >
+        ARTBOT_MAX_CHARS_RESPONSE
```
grantoesterling, line 236: Is this just an arbitrary limit or is it a limitation of the Discord embed? if it is the latter we'd need to account for the "NOTE: I am still in beta, my answers may be wrong." and "Source Documents:" text below too

src/ProjectConfig/channels.json
```diff
@@ -320,6 +320,9 @@
       }
     }
   },
+  "1103051072756519012": {
+    "name": "Server: Art Blocks Inc (822311470133542912), CH: #artgpt"
```
grantoesterling, line 324: Helpful name!!!! this resolves my other comment above i think

src/Utils/ArtGPT/dataIngestor.ts
```diff
@@ -0,0 +1,142 @@
...
+//         or import as a module and call `ingest([<GH_REPO_URLs>])`
+///////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Default GH repos to ingest if not explicitly specified
+const GH_API_BASE_URI = 'https://api.github.com/repos/'
+const GH_REPOS = ['ArtBlocks/artblocks-docs', 'ArtBlocks/artblocks-contracts']
+const GH_FILETYPES = ['.sol', '.md']
+const GH_REQ_OPTIONS = {
+  headers: {
+    Authorization: `token ${process.env.GITHUB_TOKEN}`,
```
grantoesterling, line 24: I believe this is missing from the .env.example no? could you add it in there with a note of usage (so people know it is not necessary)?

grantoesterling (approved):
LGTM (conditional ofc on yall adding the env vars to Heroku) - a few nits

this is so sick - great work!

src/Classes/ArtGPTBot.ts
jakerockland: Good catch–fixed! 🙏

src/Classes/ArtGPTBot.ts
jakerockland: Thank you for flagging! This is fixed now 🙏

src/Classes/ArtGPTBot.ts
jakerockland: Yeah that's a good point – we definitely could add this information to be exposed via `projectConfig.ts` rather than being inline directly, is that your suggestion/preference @grantoesterling ? I don't have a strong opinion on this.

src/Classes/ArtGPTBot.ts
jakerockland: Updated 🙏

src/ProjectConfig/channels.json
jakerockland: 🙏 🙌

src/Classes/ArtGPTBot.ts
jakerockland: IIUC the embed limit is `4096` for the description field, so I set it to `4000` to give buffer for these additional pieces of text. 

We could take a more robust approach here, but I think this is safe while we're experimenting given that we currently use a static `k` (number of documents) and have static "helper text", so my inclination is that this "estimate approach" is probably fine, WDYT ser?

I did just push a change that makes this a bit clearer documentation-wise.

src/Utils/ArtGPT/dataIngestor.ts
jakerockland: Good call! Updated with a comment inline explaining the field and that it is currently only needed for manual backfills.

src/Classes/ArtGPTBot.ts
grantoesterling: yep definitely makes sense to me! just wanted to make sure they were accounted for :)

src/Classes/ArtGPTBot.ts
```diff
@@ -0,0 +1,309 @@
...
+      ],
+    })
+    this.pineconeClient = new PineconeClient()
+    this.initializeLangchain()
+  }
+
+  /**
+   * Helper to initialize langchain setup.
+   */
+  async initializeLangchain() {
```
grantoesterling, line 82: Hmm i think an essential part of this is that people should be able to run/develop artbot still without needing to have these new GPT env vars set. Looking at the deployment, it seems langchain is throwing an error due to no OpenAI API key and crashing the whole program

Can you please make sure this fails gracefully when no env vars are set?

PR #505, opened 2023-05-02
Fixing engine links

grantoesterling:
Using generator_url as a fallback when there is no external_url set in the token api (occasionally happens for new engine projects)

### DIFF
--- a/src/Classes/APIBots/ReservoirListBot.ts
+++ b/src/Classes/APIBots/ReservoirListBot.ts
@@ -174,7 +174,7 @@ export class ReservoirListBot extends APIPollBot {
       {
         name: 'Live Script',
         value: `[view on artblocks.io](${
-          (!artBlocksData.external_url ?? artBlocksData.generator_url) +
+          (artBlocksData.external_url ?? artBlocksData.generator_url) +
           LISTING_UTM
         })`,
         inline: true,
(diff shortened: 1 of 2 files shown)


src/Classes/APIBots/ReservoirListBot.ts
```diff
@@ -174,7 +174,8 @@ export class ReservoirListBot extends APIPollBot {
       {
         name: 'Live Script',
         value: `[view on artblocks.io](${
-          artBlocksData.external_url + LISTING_UTM
+          (!artBlocksData.external_url ?? artBlocksData.generator_url) +
```
asaltste, line 177: nit: I don't think u need the ! here

src/Classes/ProjectBot.ts
```diff
@@ -167,12 +167,10 @@ export class ProjectBot {
     )
     const artBlocksData = artBlocksResponse.data
 
-    const osUrl = `https://api.opensea.io/api/v1/asset/${this.coreContract}/${tokenID}/`
-
     const titleLink =
       artBlocksData.external_url !== ''
```
asaltste, line 171: nit: should we be consistent here with the ??

src/Classes/APIBots/ReservoirListBot.ts
grantoesterling: oops good catch!

asaltste (approved):
some nits

ryley-o (approved):

PR #541, opened 2023-07-17
Chinese thank you Grant support

andrewjiang:
Added Chinese language support for traditional, simplified, and pinyin Chinese


asaltste (approved):
Put all 3 through google translate and output was "thanks". Thanks, [andrewjiang](https://github.com/andrewjiang)!

PR #545, opened 2023-07-20
Adding tags!

grantoesterling:
Adding tags to artbot! Heritage info is in tags so removing that old handling

`#? animated` and `#? granto.eth animated` both work!

Plus adding `yarn dev` to the package.json, bc of context switching 🙃


src/Classes/APIBots/utils.ts
```diff
@@ -118,25 +117,21 @@ async function getOSName(address: string): Promise<string> {
...
 
-function isWallet(msg: string): boolean {
+export function isWallet(msg: string): boolean {
   return msg.startsWith('0x') || msg.endsWith('eth')
 }
 
-function isVerticalName(msg: string): boolean {
+export function isVerticalName(msg: string): boolean {
   return (
     msg === 'curated' ||
```
asaltste, line 126: nit: would be nice for a bit of refactoring work if this was just an array of strings and then used one of the built in array function to see if the message equaled any of the array values

asaltste: could also make it a switch with fall through

src/Classes/ArtIndexerBot.ts
```diff
@@ -72,21 +74,21 @@ export class ArtIndexerBot {
...
         `ArtIndexerBot: Building ${projects.length} ProjectBots using: ${this.projectFetch.name}`
       )
       for (let i = 0; i < projects.length; i++) {
         const project = projects[i]
         if (project.invocations === '0') continue
         let bday = project.start_datetime
+        // Only AB projects use vertical names. Other projects (Engine, Collabs, etc) should use the category name
         const collection = this.toProjectKey(
-          collections[`${project.contract_address}-${project.project_id}`]
+          project.vertical?.category_name?.toLowerCase() !== 'collections'
```
asaltste, line 86: nit: is there an enum value we can use here for the collections string?

src/Data/generated/graphql.ts
```diff
@@ -10848,23 +10848,23 @@ export type Webflow_Spectrum_Articles_Stream_Cursor_Value_Input = {
   webflow_item_id?: InputMaybe<Scalars['String']>;
```
asaltste, line 10848: I think we want the generated folder included in the .gitignore unless we need it for the discord bot to build.

asaltste (approved):
Some nits about refactor work but lgtm

mchrupcala (approved):

src/Data/generated/graphql.ts
asaltste: ahh cool thanks for the context

PR #548, opened 2023-07-25
Trivia

grantoesterling:
Initial triviabot implementation (not active yet, and still need to add more stuff. This is just the initial PR)


src/Classes/ArtIndexerBot.ts
```diff
@@ -248,6 +250,22 @@ export class ArtIndexerBot {
     }, 1 * 60000)
   }
 
+  async startTriviaRoutine() {
+    setInterval(() => {
+      console.log("It's trivia time!")
```
asaltste, line 255: nit: extra console log left in?

src/Classes/ArtIndexerBot.ts
```diff
@@ -248,6 +250,22 @@ export class ArtIndexerBot {
...
+        const keys = Object.keys(this.projects)
+        const projectKey = keys[Math.floor(Math.random() * keys.length)]
+        const projBot = this.projects[projectKey]
+        if (projBot && projBot.editionSize > 1 && projBot.projectActive) {
+          triviaBot.askTriviaQuestion(projBot)
+          return
+        }
+        attempts++
+      }
+    }, 1 * 60000)
```
asaltste, line 267: nit: 1 * not necessary here

src/Classes/TriviaBot.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+    const question = await this.model.call(
+      `Generate a short, cryptic, poetic, vague, difficult riddle that has the answer: "${project.projectName}". It is VERY important that you DO NOT include the answer in your response. The project description is: "${project.description}"`
+    )
+    const embed = new EmbedBuilder()
+      .setTitle('Artbot Trivia Hour')
+      .setDescription(question)
+      .setColor(randomColor())
+      .setFooter({
+        text: 'Answer by using the #? command',
```
asaltste, line 86: nit maybe we can make the answer text a const at top of file as its used in multiple places

asaltste (approved):

src/Classes/ArtIndexerBot.ts
grantoesterling: both this and the below are for debugging which will still be needed atm - will be removed in final PR

PR #549, opened 2023-07-26
Adding persistent database!

grantoesterling:
Adding persistent database to keep track of scores!

Using Supabase for the PoC, as it will be free for our use case and has a nice easy js library

### DIFF
--- a/src/Classes/TriviaBot.ts
+++ b/src/Classes/TriviaBot.ts
@@ -7,12 +7,14 @@ import { projectConfig } from '..'
 import { randomColor } from '../Utils/smartBotResponse'
 import axios from 'axios'
 import { getTokenApiUrl, replaceVideoWithGIF } from './APIBots/utils'
+import { createClient, SupabaseClient } from '@supabase/supabase-js'
 
 const CHANNEL_BLOCK_TALK = projectConfig.chIdByName['block-talk']
 export class TriviaBot {
   bot: Client
   model?: OpenAIChat
-  channel: TextChannel
+  channel?: TextChannel
+  supabaseClient?: SupabaseClient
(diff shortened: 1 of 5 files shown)


src/Classes/TriviaBot.ts
```diff
@@ -98,15 +111,86 @@ export class TriviaBot {
...
+        .from(process.env.TRIVIA_TABLE ?? '')
+        .select(`score`)
+        .eq('user', `${msg.author.username}`)
+
+      let score = 1
+      if (data?.length) {
+        score = parseInt(data[0].score) + 1
+      }
+
+      msg.reply(
```
asaltste, line 141: Does it make sense to not say the success message until the score is successfuly saved? Also just want to make sure it is possible to send multiple replies?

asaltste (approved):
LGTM a couple questions about ordering

src/Classes/TriviaBot.ts
```diff
@@ -98,15 +111,86 @@ export class TriviaBot {
...
+  async leaderboard(msg: Message) {
+    if (!this.supabaseClient) {
+      console.log("Can't get leaderboard - no Supabase API key")
+      msg.reply(
+        `Uh-oh, looks like I can't query for the leaderboard. Pester Grant until he fixes it`
+      )
+      return
+    }
+
+    const { data } = await this.supabaseClient
```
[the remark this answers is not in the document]
grantoesterling: definitely - forgot about that!

src/Classes/TriviaBot.ts
grantoesterling: Def possible, and good note

PR #553, opened 2023-07-31
Twitter

grantoesterling:
BIG Twitter PR. Still some work to do, but figured I'd break it off here so as to not make the PR even bigger


src/Classes/TwitterBot.ts
```diff
@@ -1,31 +1,157 @@
...
+      console.log('Starting Twitter listener')
+      // TODO: Uncomment this when we're ready to start listening!
+      //this.startSearchAndReplyRoutine()
+    }
+  }
+
+  async startSearchAndReplyRoutine() {
+    setInterval(() => {
+      this.search()
+    }, 1 * 20000)
```
Asupkay, line 51: Can you pull this into a const like SEARCH_INTERVAL_IN_MILLISECONDS?

Asupkay (approved):

src/Classes/TwitterBot.ts
asaltste: also why 1 * 20000 isn't 1 * not needed?

src/Classes/ProjectBot.ts
```diff
@@ -190,9 +189,51 @@ export class ProjectBot {
...
+    if (!num) {
+      console.log(`Regex not matched :(`)
+      return
+    }
+    const afterTheHash = num[0].substring(1)
+    let pieceNumber
+    if (afterTheHash[0] == '?') {
+      pieceNumber = Math.floor(Math.random() * this.editionSize)
+    } else {
+      pieceNumber = parseInt(afterTheHash)
```
asaltste, line 211: doesn't this throw an uncaught error if the number is bad? will this crash the bot?

src/Classes/TwitterBot.ts
```diff
@@ -1,31 +1,158 @@
...
+    if (!tokenId) {
+      console.error(`No token found for ${projectKey}`)
+      return
+    }
+
+    const { data: artBlocksData } = await axios.get(
+      getTokenApiUrl(projectBot.coreContract, tokenId)
+    )
+
+    // TODO: support gifs
```
asaltste, line 107: nit: add the ticket to the todos

asaltste (approved):

PR #557, opened 2023-08-02
Adding UTM and gif/mp4 support

grantoesterling:
- Adding UTM and gif/mp4 support
- Adding Twitter wallet query handling
- Improving retries / robustness

### DIFF
--- a/src/Classes/APIBots/utils.ts
+++ b/src/Classes/APIBots/utils.ts
@@ -123,7 +123,7 @@ export async function getOSName(address: string): Promise<string> {
 }
 
 export function isWallet(msg: string): boolean {
-  return msg.startsWith('0x') || msg.endsWith('eth')
+  return !!msg.match(/(0x[a-fA-F0-9]{40})|([a-zA-Z0-9.-]+\.eth)/g)
 }
 
 const acceptedVerticals = [
--- a/src/Classes/APIBots/utils.ts
+++ b/src/Classes/APIBots/utils.ts
@@ -28,11 +28,16 @@ const osAddressMap: { [id: string]: string } = {}
 const MAX_ENS_RETRIES = 3
 
 // UTM for links so we can track traffic that comes through Artbot
-const ARTBOT_UTM = '?utm_source=artbot&utm_medium=discord'
-export const LISTING_UTM = ARTBOT_UTM + '&utm_campaign=listing'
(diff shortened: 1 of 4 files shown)


src/Classes/APIBots/utils.ts
```diff
@@ -28,11 +28,16 @@ const osAddressMap: { [id: string]: string } = {}
...
 // UTM for links so we can track traffic that comes through Artbot
-const ARTBOT_UTM = '?utm_source=artbot&utm_medium=discord'
-export const LISTING_UTM = ARTBOT_UTM + '&utm_campaign=listing'
-export const SALE_UTM = ARTBOT_UTM + '&utm_campaign=sale'
-export const MINT_UTM = ARTBOT_UTM + '&utm_campaign=mint'
-export const PROJECTBOT_UTM = ARTBOT_UTM + '&utm_campaign=projectbot'
+const ARTBOT_UTM = '?utm_source=artbot'
+const DISCORD_UTM = '&utm_medium=discord'
+const TWITTER_UTM = '&utm_medium=twitter'
+export const LISTING_UTM = ARTBOT_UTM + DISCORD_UTM + '&utm_campaign=listing'
```
asaltste, line 34: nit: I think we can use `` instead of + to construct these strings but nbd

Asupkay (approved):

src/Classes/ArtIndexerBot.ts
```diff
@@ -231,35 +237,64 @@ export class ArtIndexerBot {
...
+    let afterTheHash = content.replace(/#(\?|\d+)/g, '').trim()
+    let key = this.toProjectKey(afterTheHash)
+    key = content === '#?' ? '#?' : key
 
-    const messageType = this.getMessageType(projectKey, afterTheHash)
+    const messageType = this.getMessageType(key, afterTheHash)
+    let projectBot
     if (messageType === MessageTypes.WALLET) {
-      // TODO: support wallet tweets https://app.asana.com/0/1201568815538912/1205173752010557/f
+      const walletMatch = afterTheHash.match(
```
asaltste, line 269: can we use the isWallet function here

grantoesterling: nah bc we need to get the actual wallet string (which is what match does)

asaltste (approved):

PR #568, opened 2023-08-09
Timer

grantoesterling:
Adding Scheduler class that uses cron to manage time based messages :)


src/Classes/SchedulerBot.ts
```diff
@@ -0,0 +1,85 @@
+import { Channel, Collection, EmbedBuilder, TextChannel } from 'discord.js'
+import { ProjectConfig } from '../ProjectConfig/projectConfig'
+import { artIndexerBot } from '..'
+import { delay } from './APIBots/utils'
+import { randomColor } from '../Utils/smartBotResponse'
+
+const cron = require('node-cron')
```
asaltste, line 7: nit: should this be import cron from 'node-cron'

src/Classes/SchedulerBot.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  async initialize() {
+    await delay(8000) // Wait for bot to connect and channels to load
+    console.log('Starting Scheduler')
+    const channels = this.channels
+    const projectConfig = this.projectConfig
+
+    // Birthdays
+    cron.schedule(
+      '0 1,9,17 * * *',
```
asaltste, line 29: nit: add a comment about what this time is

src/Classes/SchedulerBot.ts
```diff
@@ -0,0 +1,85 @@
...
+      },
+      null,
+      true,
+      'America/Chicago'
+    )
+
+    // Marfa
+    const sendMarfaMessage = this.sendMarfaMessage
+    cron.schedule(
+      '0 11 * * *',
```
asaltste, line 42: nit: add a comment about what this time is

asaltste (approved):

mchrupcala (approved):

src/Classes/SchedulerBot.ts
grantoesterling: thats what i wanted to do, but for reasons I can't comprehend, the node-cron library only works this way :(

PR #570, opened 2023-08-10
swapping cron libraries

grantoesterling:
Swapping cron libraries bc node-cron is apparently ancient

Also adding temporary debugging logging to see when it thinks the next events are


package.json
```diff
@@ -30,6 +30,7 @@
     "axios": "^1.1.3",
     "body-parser": "^1.15.2",
     "chatgpt": "^5.2.5",
+    "croner": "^6.0.7",
```
Asupkay, line 33: It's the driving croner baby

Asupkay (approved):

PR #579, opened 2023-08-22
Recent mints from different contracts

asaltste:
This is a pr to add the ability to get the most recent minted token on a per contract basis. 

Commands:
`#recent` //gets the most recent minted flagship token
`#recent <contract_address>` //gets the most recent minted token for that particular contract
`#recent <contract_name>` //gets the most recent token minted for the particular named contract
`#recent <alias>` //gets the most recent minted token for that series of contracts associated with the alias

---
  - https://app.asana.com/0/0/1205326334812414


src/Data/queryGraphQL.ts
```diff
@@ -428,3 +432,31 @@ export async function getProjectFloor(projectId: string) {
     return undefined
   }
 }
+
+export async function getMostRecentMintedTokenByContracts(
+  contracts: string[]
+): Promise<ProjectTokenDetailFragment> {
+  const { data } = await client
+    .query(GetMostRecentMintedTokenByContractDocument, {
```
grantoesterling, line 440: Should use: `const hasuraClient = getClientForContract(...)` in someway to handle prohibition / arbitrum case

src/Classes/ArtIndexerBot.ts
```diff
@@ -237,13 +272,16 @@ export class ArtIndexerBot {
       msg.content = `#${token?.invocation}`
       projectBot = this.projects[this.toProjectKey(token.project.name ?? '')]
     } else {
-      projectBot = await this.projectBotForMessage(projectKey, afterTheHash)
+      if (!projectBot) {
```
grantoesterling, line 279: undo dis

src/Classes/ArtIndexerBot.ts
```diff
@@ -210,11 +237,19 @@ export class ArtIndexerBot {
...
 
-    const messageType = this.getMessageType(projectKey, afterTheHash)
+    const messageType = this.getMessageType(
+      projectKey,
+      afterTheHash,
+      msg.content
+    )
 
     let projectBot
+    if (messageType === MessageTypes.RECENT) {
```
grantoesterling, line 251: nit: could this be an elseif with the Wallet if below

src/Classes/ArtIndexerBot.ts
```diff
@@ -55,12 +59,16 @@ export class ArtIndexerBot {
...
   tags: { [id: string]: ProjectBot[] }
+  idProjects: { [id: string]: ProjectBot }
+  contracts: { [id: string]: ContractDetailFragment }
   walletTokens: { [id: string]: TokenDetailFragment[] }
   initialized = false
 
   constructor(projectFetch = getAllProjects) {
     this.projectFetch = projectFetch
     this.projects = {}
+    this.idProjects = {}
```
grantoesterling, line 70: literally the nittiest nit but could this be "projectsById" or something more idiomatic

src/Classes/ArtIndexerBot.ts
```diff
@@ -73,6 +81,7 @@ export class ArtIndexerBot {
    * Initialize async aspects of the FactoryBot
    */
   async init() {
+    await this.buildContracts()
```
grantoesterling, line 84: probs want to put this in the `if (this.projectFetch === getAllProjects)` statement below so that it is only run for the main instance of this bot

src/Classes/ArtIndexerBot.ts
```diff
@@ -86,6 +95,16 @@ export class ArtIndexerBot {
     }, parseInt(METADATA_REFRESH_INTERVAL_MINUTES) * ONE_MINUTE_IN_MS)
   }
 
+  async buildContracts() {
+    const contractArr = await getAllContracts()
+    for (let i = 0; i < contractArr.length; i++) {
+      const name = contractArr[i].name
```
grantoesterling: i wanna think more about this, but we might want to be using the `.toProjectKey`ified version of the name as the key instead of the actual name, so that it is easier to index by later (ie the discord user doesn't have to get the capitalization and punctuation right)

src/Data/queryGraphQL.ts
asaltste: Ok I dont think the named thing will work then tho. Cause then the client in getAllContracts wont work

src/Classes/ArtIndexerBot.ts
```diff
@@ -210,7 +248,11 @@ export class ArtIndexerBot {
 
     let projectKey = this.toProjectKey(afterTheHash)
 
-    const messageType = this.getMessageType(projectKey, afterTheHash)
+    const messageType = this.getMessageType(
+      projectKey,
+      afterTheHash,
+      msg.content
```
grantoesterling, line 254: probably want to .toLowerCase this so `#Recent` still works

src/Data/queryGraphQL.ts
```diff
@@ -428,3 +433,45 @@ export async function getProjectFloor(projectId: string) {
...
+    throw Error(
+      'No data returned from getMostRecentMintedFlagshipToken Hasura query'
+    )
+  }
+
+  return data.tokens_metadata[0]
+}
+
+export async function getAllContracts(): Promise<ContractDetailFragment[]> {
+  const { data } = await client.query(GetAllContractsDocument, {}).toPromise()
```
grantoesterling, line 470: Good point above - you'll need to do two separate queries with the two different clients and then concat them - rn Prohibition isn't getting pulled in here

Arbitrum throws such a wrench into this stuff

src/Data/queryGraphQL.ts
grantoesterling: see comment below

src/Data/queryGraphQL.ts
```diff
@@ -428,3 +433,45 @@ export async function getProjectFloor(projectId: string) {
...
+    throw Error(
+      'No data returned from getMostRecentMintedTokenByContracts Hasura query'
+    )
+  }
+
+  return data.tokens_metadata[0]
+}
+
+export async function getMostRecentMintedFlagshipToken(): Promise<ProjectTokenDetailFragment> {
+  const { data } = await client
```
grantoesterling, line 456: Can't this just call `getMostRecentMintedTokenByContracts` using core contracts as args instead of being its own query?

src/Classes/ArtIndexerBot.ts
```diff
@@ -316,6 +362,56 @@ export class ArtIndexerBot {
...
+      }
+      const token = await getMostRecentMintedTokenByContracts(contracts)
+      return token
+    } catch (e) {
+      console.error('Error in getContractTokenForKey', e)
+      return null
+    }
+  }
+
+  async getRecentProjectBot(key: string, message: Message) {
```
grantoesterling, line 394: nit: i think this fxn should be called something else like `sendRecentMessage` or something, since it isn't really "getting" a bot

src/Data/queryGraphQL.ts
asaltste: I don't want to have to maintain the list of the flagship contract so I am instead using the vertical way to do this

src/Classes/ArtIndexerBot.ts
asaltste: cool will change that :) good point

asaltste: I do think its nice to have this in a function so that it can be called from tweets and the discord tbf.

src/Data/queryGraphQL.ts
grantoesterling: The core contracts list already exists and is used multiple times in this file, though I agree we should move away from having a hardcoded list we have to maintain. I'll make a ticket for changing the way that works, but in the meantime can this function use that list instead of this for consistency?

src/Classes/ArtIndexerBot.ts
```diff
@@ -236,6 +286,24 @@ export class ArtIndexerBot {
...
+          // use flagship contract
+          token = await getMostRecentMintedFlagshipToken()
+        }
+
+        const projectId = token.project_id
+
+        const projectBot = this.projectsById[projectId]
+        msg.content = `#${token?.invocation}`
+
+        projectBot.handleNumberMessage(msg)
```
grantoesterling, line 302: pretty sure you don't need to return early anymore - you can get rid of the try catch and this line, and the projectbot will be called in the last line of the function as it is for all other queries

grantoesterling: I think the try catch can get moved to the whole block calling `handleNumberMessage` in index.ts

src/Data/queryGraphQL.ts
asaltste: I'd rather just ensure its this way rather than doing consistency sake here. Cause I know this is how it should be done and rather keep this as is then use the existing bad way

asaltste: Just cause this is the way artblocks-viewer does it as well which is what this is trying to emulate

grantoesterling: Where I'm coming from: Artbot used to have just a ton of different subgraph and API calls, some accomplishing the same thing bc they were written by different people who weren't aware of other parts of the codebase. Part of my efforts the past year have been to cull those things and commonize the methods as much as I can. 

This may technically be the correct way to do it, but by using the existing CORE_CONTRACTS, it becomes much easier for whoever does the refactoring work in the future to include this as part of the refactor.

I think its fine to merge it the way it is, this isn't really a hill i want to die on lol i just like having stuff consistent haha

grantoesterling (approved):

src/Data/queryGraphQL.ts
asaltste: Yeah I 100% feel u I think just gonna leave as is for now and we can revisit with a further ticket if thats all good

PR #580, opened 2023-08-23
don't show price for flagged tokens

grantoesterling:
Exclude flagged tokens from #floor calls and don't show price if flagged


src/Data/graphql/artbot-hasura-queries.graphql
```diff
@@ -132,9 +133,10 @@ query getProjectInvocations($id: String!) {
 
 query getProjectFloor($id: String!) {
   projects_metadata(where: { id: { _eq: $id } }) {
-    tokens(limit: 1, order_by: { list_eth_price: asc }) {
+    tokens(limit: 3, order_by: { list_eth_price: asc }) {
       invocation
       list_eth_price
+      isFlaggedAsSuspicious
```
Asupkay, line 139: smart move no adding it as a filter parameter. We should have a ticket blocked on moving over the isFlagged field to reservoir to know that we should switch this.

src/Data/queryGraphQL.ts
```diff
@@ -394,9 +394,9 @@ export async function getProjectFloor(projectId: string) {
     if (!data) {
       throw Error('No data returned from getProjectFloor Hasura query')
     }
-    return data.projects_metadata.length > 0
-      ? data.projects_metadata[0].tokens?.[0]
-      : null
+    return data.projects_metadata[0].tokens.find(
+      (token) => !token.isFlaggedAsSuspicious
```
Asupkay, line 398: What happens if all 3 tokens are flagged?

grantoesterling: just nothing is responded - i think this is a rare enough case we can ignore it for now

src/Data/graphql/artbot-hasura-queries.graphql
grantoesterling: ticket created!

PR #584, opened 2023-08-28
adding identical listing catching

grantoesterling:
Catching identical listings by saving a local dict of tokenId -> last listed price. If new listing for that token is within 0.0001 of that previous listing, don't post it!

Also, updating the Reservoir endpoint to v5, and switching over to using their library for some typing. Not necessary per se, but just wanted to do it while im in here

### DIFF
--- a/src/Classes/APIBots/ReservoirListBot.ts
+++ b/src/Classes/APIBots/ReservoirListBot.ts
@@ -5,7 +5,7 @@ import {
   sendEmbedToListChannels,
 } from '../../Utils/activityTriager'
 import { APIPollBot } from './ApiPollBot'
-
+import { paths } from '@reservoir0x/reservoir-sdk'
 import {
   LISTING_UTM,
   ensOrAddress,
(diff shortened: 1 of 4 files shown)


src/Classes/APIBots/ReservoirListBot.ts
```diff
@@ -99,11 +102,21 @@ export class ReservoirListBot extends APIPollBot {
...
+    const price = listing.price?.amount?.decimal ?? 0
+    const currency = listing.price?.currency?.symbol
     const owner = listing.maker
-    let platform = listing.source.name
-    if (listing.source.domain.includes('artblocks')) {
+    let platform = listing.source?.name
+
+    if (
+      this.recentListings[tokenID] &&
+      Math.abs(this.recentListings[tokenID] - price) < IDENTICAL_TOLERANCE
```
asaltste, line 112: I think we want this to be a <=

asaltste (approved):

src/Classes/APIBots/ReservoirListBot.ts
grantoesterling: truuuu

PR #592, opened 2023-09-14
collision fixes

grantoesterling:
Fixing collision problem, plus some extras:
- All #?,  #? art blocks, or generic Artbot calls only resolve to what is on the AB website (curated, presents, explorations, collaborations)
- Use #? all to call AB projects + Engine
- For project collisions – AB projects take precedence
- To call non-AB projects that have a collision – #? [artist name] [project]

- Also, adding `#? <engine platform>` command


Asupkay (approved):

src/Classes/ArtIndexerBot.ts
```diff
@@ -182,7 +196,34 @@ export class ArtIndexerBot {
...
+          const platform = this.toProjectKey(
+            project.contract.name ?? 'Unknown contract'
+          )
+          this.platforms[platform] = this.platforms[platform] ?? []
+          this.platforms[platform].push(newBot)
+        }
       }
+
+      // Set up contract aliases
+      CONTRACT_ALIASES.forEach((item) => {
```
mchrupcala, line 214: Could we try this without the nested forEach? Partner contracts are a small list now but as the list grows this 3x nesting feels like it could add complexity over time.

mchrupcala (approved):

PR #627, opened 2023-12-04
Add spin alias

pmohun:
SPINᵗ has a special superscript character which makes it hard to easily call up in #block-talk. Adding an alias for "Spin" so it's easier for people to enjoy Matt's great art.


grantoesterling (approved):

pmohun:
thanks grant!

PR #637, opened 2024-01-07
Replace projectName with underscores using a case-insensitive global regular expression

dols-codes:
prevents artbot from including project name in trivia questions


grantoesterling (approved):
LEGEND

PR #698, opened 2024-11-20
trivia season 3

grantoesterling:


src/Data/supabase.ts
```diff
@@ -26,7 +26,11 @@ export const updateTriviaScore = async (username: string): Promise<number> => {
   // NOTE: When changing seasons, we have to manually change the upsert column here. Super annoying I know.
   const { error } = await supabaseClient
     .from(process.env.TRIVIA_TABLE ?? '')
-    .upsert({ user: `${username}`, score: totalScore, season_two: seasonScore })
+    .upsert({
+      user: `${username}`,
+      score: totalScore,
+      season_three: seasonScore,
```
asaltste, line 32: nit: you can do [CURRENT_SEASON]:seasonScore  then you don't need to manually change this with every season change

asaltste (approved):

src/Data/supabase.ts
asaltste: it's how u make an object param be a variable

grantoesterling: hmm i dont think it let me do this when i created this but lemme try again

PR #723, opened 2025-08-27
using opensea api for lists/sales

grantoesterling:
Switching to the [OpenSea Stream API](https://github.com/ProjectOpenSea/stream-js) for sale/list events! 

There is no contract level filtering on OS end, so it seems the only way is just to receive all sale/list events and manually filter by contract on our side

Sale bot is hooked up to Twitter so it all should work in theory!

Keeping the Reservoir code around in case we need to revert this, but i will open a ticket to remove it all once it is proven stable

### DIFF
--- a/src/Classes/APIBots/OpenSeaListBot.ts
+++ b/src/Classes/APIBots/OpenSeaListBot.ts
@@ -1,0 +1,227 @@
+import axios from 'axios'
+import { Client, ColorResolvable, EmbedBuilder } from 'discord.js'
+import {
+  BAN_ADDRESSES,
+  sendEmbedToListChannels,
+} from '../../Utils/activityTriager'
+import {
+  LISTING_UTM,
+  ensOrAddress,
+  getCollectionType,
+  getTokenApiUrl,
+  getTokenUrl,
+  replaceVideoWithGIF,
+  isEngineContract,
+  isExplorationsContract,
+} from './utils'
+import { ItemListedEvent } from '@opensea/stream-js'
+
+const IDENTICAL_TOLERANCE = 0.0001
+const LISTING_TTL_MS = 24 * 60 * 60 * 1000 // 24 hours
+
+/**
+ * Bot for handling OpenSea stream listing events
+ * Similar to ReservoirListBot but for OpenSea stream data
+ */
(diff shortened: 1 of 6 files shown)


src/Classes/APIBots/OpenSeaListBot.ts
```diff
@@ -0,0 +1,227 @@
...
+  /**
+   * Parses OpenSea NFT ID and extracts contract address and token ID
+   * @param nftId - Format: "ethereum/0x2308742aa28cc460522ff855d24a365f99deba7b/7111"
+   * @returns {contractAddress, tokenId} or null if invalid format
+   */
+  private parseNftId(
+    nftId: string
+  ): { contractAddress: string; tokenId: string } | null {
+    const parts = nftId.split('/')
+    if (parts.length !== 3) {
```
ryley-o, line 75: should we filter to only ethereum?

src/index.ts
```diff
@@ -394,11 +393,103 @@ if (!isTwitterEnabled) {
...
+
+// Helper function to check if an OpenSea NFT ID contains any of our tracked contracts
+const isTrackedContract = (
+  nftId: string,
+  contractsArray: string[]
+): boolean => {
+  // NFT ID format: "ethereum/0x2308742aa28cc460522ff855d24a365f99deba7b/7111"
+  // Extract the contract address (middle part)
+  const parts = nftId.split('/')
+  if (parts.length !== 3) return false
```
ryley-o, line 420: same comment re: filter to ethereum?

ryley-o (approved):

src/index.ts
```diff
@@ -312,68 +311,68 @@ discordClient.on(Events.MessageCreate, async (msg) => {
...
-      {
-        Accept: 'application/json',
-        'x-api-key': process.env.RESERVOIR_API_KEY,
-      },
-      abTwitterBot
-    )
-    botsToCleanup.push(saleBot)
-  }
-}
+// const initReservoirBots = async () => {
```
lindsgil, line 314: nit: can we add a comment here around reservoir deprecation + the move to OS 🙏

lindsgil (approved):
🔥 🔥 🔥

